PR #6, opened 2020-03-11
Upgrade Kibana to 7.4.2

yizheliu-amazon:
*Description of changes:*
Upgrade Kibana to 7.4.2. Make sure all the functionalities(create/edit/delete detector/feature) are working as before, and UI looks same as before. 


package.json
```diff
@@ -1,20 +1,20 @@
 {
   "name": "opendistro-anomaly-detection",
-  "version": "1.2.1.0",
+  "version": "1.4.2.0",
```
ylwu-amzn: Why use "1.4.2.0",  I see alerting [package.json](https://github.com/opendistro-for-elasticsearch/alerting-kibana-plugin/blob/opendistro-1.4/package.json) using "1.4.0.0"

yizheliu-amazon: I saw our previous version is 1.2.1, which matches to old Kibana version 7.2.1, that is why I use 1.4.2 in new version, as our new Kibana version is 7.4.2. I guess that is our convention. On the other side, Alerting always uses 1.2.0.0, 1.3.0.0, 1.4.0.0 for 7.2.0, 7.3.2, 7.4.2, it looks like that is their [convention](https://github.com/opendistro-for-elasticsearch/alerting-kibana-plugin/commit/e48ac68a58c1fe8a82e3386c88fe2e4fafce75ec#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L3-R3). I have no preference for version number. Feel free to let me know which I should use.

public/pages/PreviewDetector/containers/AdjustModel.tsx
```diff
@@ -90,6 +90,12 @@ export function AdjustModel(props: AdjustModelProps) {
       );
     }
   };
+  const flyoutStyleTop = props.isSticky ? '111px' : `${224 - props.topOffset}px`;
+  const flyoutStyle = {
+    top: flyoutStyleTop,
+    borderTop: '0px',
+    height: 'calc(100% - ' + flyoutStyleTop +')',
```
mihirsoni: You can use string litterate similar to line 93
```
height: `calc(100%-${flyoutstypeTop})`
```

yizheliu-amazon: thanks. Have used this way.

Issue #11, opened 2020-03-25
Start/pause AD job


ylwu-amzn:
Anomaly detection ES plugin has its own job, so we need to support starting/stopping AD job

PR #12, opened 2020-03-25
add start/stop AD job api

ylwu-amzn:
*Issue #11 

*Description of changes:*
Add start/stop AD job apis

### DIFF
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -55,8 +55,8 @@ export default function (apiRouter: Router) {
   apiRouter.post('/detectors/{detectorId}/preview', previewDetector);
   apiRouter.get('/detectors/{detectorId}/results', getAnomalyResults);
   apiRouter.delete('/detectors/{detectorId}', deleteDetector);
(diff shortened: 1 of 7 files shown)


server/routes/ad.ts
```diff
@@ -55,6 +55,8 @@ export default function (apiRouter: Router) {
   apiRouter.post('/detectors/{detectorId}/preview', previewDetector);
   apiRouter.get('/detectors/{detectorId}/results', getAnomalyResults);
   apiRouter.delete('/detectors/{detectorId}', deleteDetector);
+  apiRouter.post('/detectors/{detectorId}/_start', startDetector);
```
mihirsoni: np: I would not recommend to add `_` let's be consistent and remove it from kibana api.

ylwu-amzn:
> What would happen if the Job is already started, do we disable on UI also how does backend handle this ?

Will disable update on UI and we have check from backend, will block updating detector if job is running.

server/routes/ad.ts
ylwu-amzn: Sure, will change it

PR #16, opened 2020-04-03
Update detector list page & add profile API

ohltyler:
*Issue #, if available:*

*Description of changes:*

Screen Shot 2020-04-02 at 11 36 40 AM

This PR modifies the previous detector list page:
- adds the profile api to pull detector state information
- adds the "Indices", "Detector state", and "Last updated" columns to the table
- adds detector state and index dropdown filtering capability
- adds search capability in the index dropdown
- fixes the bug which was preventing aggregated anomaly results from being pulled from the backend
- displays the number of selected detectors in the page title
- adds sorting capabilities by all table fields

Testing:
- adds unit tests to test indices being displayed and some of the filtering functionality
- all unit tests pass
- manual testing to confirm that filtering by detector state and index works properly

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #17, opened 2020-04-06
Add Anomaly Dashboard

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Revision 4:
Add detector/state/indices filter. State filter function will be implemented once [Profile API PR](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/16)
is merged.

Screen Shot 2020-04-14 at 3 08 24 PM

Screen Shot 2020-04-14 at 3 08 34 PM

 

Revision 3:
Add detetor/feature list, Address Mihir, Yaliang's comments

Screen Shot 2020-04-10 at 10 58 32 AM

Revision 2: addressed Tyler's comment on Query string, and add Anomaly Distribution chart. 
Screen Shot 2020-04-08 at 10 50 17 AM

Add Anomaly Live Chart for Dashboard page. 

Filter box, sunburst chart and detector/feature list will be coming later.

1 detector case(interval 10 mins):
Screen Shot 2020-04-05 at 9 59 10 PM

Most extreme case(10 detectors, 1 min interval):
Screen Shot 2020-04-04 at 6 08 30 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "@babel/preset-typescript": "^7.3.3",
     "@elastic/eslint-config-kibana": "link:../../packages/eslint-config-kibana",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/kbn-eslint-import-resolver-kibana",
-    "@elastic/eui": "6.10.7",
     "@kbn/config-schema": "link:../../packages/kbn-config-schema",
     "@kbn/plugin-helpers": "link:../../packages/kbn-plugin-helpers",
     "@testing-library/jest-dom": "^4.0.0",
@@ -71,10 +70,11 @@
     "typescript": "3.0.3"
   },
   "dependencies": {
+    "@elastic/charts": "^18.2.2",
+    "babel-polyfill": "^6.26.0",
     "formik": "^1.5.8",
     "query-string": "^6.8.2",
     "react-redux": "^7.1.0",
-    "reselect": "^4.0.0",
-    "babel-polyfill": "^6.26.0"
+    "reselect": "^4.0.0"
   }
 }
--- a/public/components/ContentPanel/ContentPanel.tsx
+++ b/public/components/ContentPanel/ContentPanel.tsx
@@ -25,6 +25,9 @@ import {
 } from '@elastic/eui';
 
 type ContentPanelProps = {
+  // keep title string part for backwards compatibility
+  // might need to refactor code and
+  // deprecate support for 'string' in the near future
   title: string | React.ReactNode | React.ReactNode[];
   titleSize?: EuiTitleSize;
   subTitle?: React.ReactNode | React.ReactNode[];
--- a/public/components/ContentPanel/index.scss
+++ b/public/components/ContentPanel/index.scss
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
@@ -13,15 +13,13 @@
  * permissions and limitations under the License.
  */
 
-import {
-  SORT_DIRECTION,
-  AD_DOC_FIELDS,
-} from '../../../../server/utils/constants';
+.content-panel-title {
+  color: #3f3f3f;
+}
 
-export const DEFAULT_QUERY_PARAMS = {
-  from: 0,
-  search: '',
-  size: 20,
-  sortDirection: SORT_DIRECTION.ASC,
-  sortField: AD_DOC_FIELDS.DATA_START_TIME,
-};
+.content-panel-subTitle {
+  color: #879196;
+  font-family: 'Helvetica Neue';
+  font-size: 12px;
+  letter-spacing: 0;
+}
--- a/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
@@ -1,0 +1,244 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import { DetectorListItem } from '../../../models/interfaces';
+import { useState, useEffect } from 'react';
+import {
+  fillOutColors,
+  visualizeAnomalyResultForSunburstChart,
+  getLatestAnomalyResultsForDetectorsByTimeRange,
+} from '../utils/utils';
+import ContentPanel from '../../../components/ContentPanel/ContentPanel';
+import {
+  EuiSelect,
+  EuiText,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiLoadingChart,
+  //@ts-ignore
+  EuiStat,
+} from '@elastic/eui';
+import { Chart, Partition, PartitionLayout } from '@elastic/charts';
+import { useDispatch } from 'react-redux';
+import { Datum } from '@elastic/charts/dist/utils/commons';
+import React from 'react';
+import { TIME_RANGE_OPTIONS } from '../../Dashboard/utils/constants';
+import { get } from 'lodash';
+import { searchES } from '../../../redux/reducers/elasticsearch';
+import { MAX_DETECTORS } from '../../../utils/constants';
+import { AD_DOC_FIELDS } from '../../../../server/utils/constants';
+export interface AnomaliesDistributionChartProps {
+  allDetectorsSelected: boolean;
+  selectedDetectors: DetectorListItem[];
+}
+
+export const AnomaliesDistributionChart = (
+  props: AnomaliesDistributionChartProps
+) => {
+  const dispatch = useDispatch();
+
+  const [anomalyResults, setAnomalyResults] = useState([] as object[]);
+
+  // TODO: try to find a better way of using redux,
+  // which can leverage redux, and also get rid of issue with multiple redux on same page,
+  // so that we don't need to manualy update loading status
+  // Issue link: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/23
+  const [anomalyResultsLoading, setAnomalyResultsLoading] = useState(true);
+  const [finalDetectors, setFinalDetectors] = useState(
+    [] as DetectorListItem[]
+  );
+
+  const [indicesNumber, setIndicesNumber] = useState(0);
+
+  const [timeRange, setTimeRange] = useState(TIME_RANGE_OPTIONS[0].value);
+
+  const getAnomalyResult = async (currentDetectors: DetectorListItem[]) => {
+    const finalAnomalyResult = await getLatestAnomalyResultsForDetectorsByTimeRange(
+      searchES,
+      props.selectedDetectors,
+      timeRange,
+      MAX_DETECTORS,
+      dispatch
+    );
+    setAnomalyResults(finalAnomalyResult);
+
+    const resultDetectors = getFinalDetectors(
+      finalAnomalyResult,
+      props.selectedDetectors
+    );
+    setIndicesNumber(getFinalIndices(resultDetectors).size);
+    setFinalDetectors(resultDetectors);
+    setAnomalyResultsLoading(false);
+  };
+
+  const getFinalIndices = (detectorList: DetectorListItem[]) => {
+    const indicesSet = new Set();
+    detectorList.forEach(detectorItem => {
+      indicesSet.add(detectorItem.indices.toString());
+    });
+
+    return indicesSet;
+  };
+
+  const getFinalDetectors = (
+    finalLiveAnomalyResult: object[],
+    detectorList: DetectorListItem[]
+  ): DetectorListItem[] => {
+    const detectorSet = new Set<string>();
+    finalLiveAnomalyResult.forEach(anomalyResult => {
+      detectorSet.add(get(anomalyResult, AD_DOC_FIELDS.DETECTOR_ID, ''));
+    });
+
+    const filteredDetectors = detectorList.filter(detector =>
+      detectorSet.has(detector.id)
+    );
+
+    return filteredDetectors;
+  };
+
+  const handleOnChange = (e: any) => {
+    setTimeRange(e.target.value);
+  };
+
+  useEffect(() => {
+    getAnomalyResult(props.selectedDetectors);
+  }, [timeRange, props.selectedDetectors]);
+
+  return (
+    <ContentPanel
+      title="Anomalies by index and detector"
+      titleSize="s"
+      subTitle={
+        <EuiFlexItem>
+          <EuiText className={'anomaly-distribution-subtitle'}>
+            <p>
+              {
+                'The inner circle shows the anomaly distribution by your indices. The outer circle shows the anomaly distribution by your detector'
+              }
+            </p>
+          </EuiText>
+        </EuiFlexItem>
+      }
+      actions={
+        <EuiSelect
+          style={{ width: 150 }}
+          id="timeRangeSelect"
+          options={TIME_RANGE_OPTIONS}
+          value={timeRange}
+          onChange={handleOnChange}
+          fullWidth
+        />
+      }
+    >
+      <EuiFlexGroup style={{ padding: '10px' }}>
+        <EuiFlexItem>
+          <EuiStat
+            description={'Indices with anomalies'}
+            title={indicesNumber}
+            isLoading={anomalyResultsLoading}
+            titleSize="s"
+          />
+        </EuiFlexItem>
+        <EuiFlexItem>
+          <EuiStat
+            description={'Detectors with anomalies'}
+            title={finalDetectors.length}
+            isLoading={anomalyResultsLoading}
+            titleSize="s"
+          />
+        </EuiFlexItem>
+      </EuiFlexGroup>
+      {anomalyResultsLoading ? (
+        <EuiFlexGroup justifyContent="center">
+          <EuiFlexItem grow={false}>
+            <EuiLoadingChart size="m" />
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiLoadingChart size="l" />
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiLoadingChart size="xl" />
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      ) : (
+        <EuiFlexGroup justifyContent="center">
+          <EuiFlexItem grow={false}>
+            <Chart className="anomalies-distribution-sunburst">
+              <Partition
+                id="Anomalies by index and detector"
+                data={visualizeAnomalyResultForSunburstChart(
+                  anomalyResults,
+                  finalDetectors
+                )}
+                valueAccessor={(d: Datum) => d.count as number}
+                valueFormatter={(d: number) => d.toString()}
+                layers={[
+                  {
+                    groupByRollup: (d: Datum) => d.indices,
+                    nodeLabel: (d: Datum) => {
+                      return d;
+                    },
+                    fillLabel: {
+                      textInvertible: false,
+                    },
+                    shape: {
+                      fillColor: d => {
+                        return fillOutColors(
+                          d,
+                          (d.x0 + d.x1) / 3 / (2 * Math.PI),
+                          []
+                        );
+                      },
+                    },
+                  },
+                  {
+                    groupByRollup: (d: Datum) => d.name,
+                    nodeLabel: (d: Datum) => {
+                      return d;
+                    },
+                    fillLabel: {
+                      textInvertible: true,
+                    },
+                    shape: {
+                      fillColor: d => {
+                        return fillOutColors(
+                          d,
+                          (d.x0 + d.x1) / 2 / (2 * Math.PI),
+                          []
+                        );
+                      },
+                    },
+                  },
+                ]}
+                config={{
+                  partitionLayout: PartitionLayout.sunburst,
+                  fontFamily: 'Arial',
+                  outerSizeRatio: 0.6,
+                  fillLabel: {
+                    textInvertible: true,
+                    fontStyle: 'italic',
+                  },
+                  // TODO: Given only 1 detector exists, the inside Index circle will have issue in following scenarios:
+                  // 1: if Linked Label is configured for identifying index, label of Index circle will be invisible;
+                  // 2: if Fill Label is configured for identifying index, label of it will be overlapped with outer Detector circle
+                  // Issue link: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/24
+                }}
+              />
+            </Chart>
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      )}
+    </ContentPanel>
+  );
+};
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -1,0 +1,267 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import { useState, useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { DetectorListItem } from '../../../models/interfaces';
+import {
+  AD_DOC_FIELDS,
+  MIN_IN_MILLI_SECS,
+} from '../../../../server/utils/constants';
+import {
+  EuiBadge,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiLoadingChart,
+  //@ts-ignore
+  EuiStat,
+} from '@elastic/eui';
+import { searchES } from '../../../redux/reducers/elasticsearch';
+import { get, isEmpty } from 'lodash';
+import moment, { Moment } from 'moment';
+import ContentPanel from '../../../components/ContentPanel/ContentPanel';
+import {
+  Chart,
+  Axis,
+  Settings,
+  Position,
+  BarSeries,
+  niceTimeFormatter,
+  ScaleType,
+  LineAnnotation,
+  AnnotationDomainTypes,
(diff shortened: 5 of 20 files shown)


public/pages/Dashboard/components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+            },
+          },
+          {
+            range: {
+              [AD_DOC_FIELDS.ANOMALY_GRADE]: {
+                gt: 0.0,
+              },
+            },
+          },
+        ],
```
ohltyler, line 92: I believe you shouldn't need the `exists` argument for the anomaly grade here. The `range` arg should be enough to only return results where anomaly grade exists AND is > 0.0. See https://github.com/ohltyler/anomaly-detection-kibana-plugin/blob/updateDetectorListPage/server/routes/utils/adHelpers.ts#L94

yizheliu-amazon: good catch, will remove it.

ohltyler: Also, small thing, but couldn't the `must_not` be moved outside of a nested `bool` and be in the outer `bool`?

yizheliu-amazon: hmmm, good idea. I agree this can simplify the query. Will change it.

package.json
```diff
@@ -33,6 +33,7 @@
     "@elastic/eslint-config-kibana": "link:../../packages/eslint-config-kibana",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/kbn-eslint-import-resolver-kibana",
     "@elastic/eui": "6.10.7",
+    "@elastic/charts": "^18.2.0",
```
ylwu-amzn, line 36: Do we plan to upgrade `@elastic/charts` or try `react vis`?

public/components/ContentPanel/ContentPanel.tsx
```diff
@@ -46,13 +51,30 @@ const ContentPanel = (props: ContentPanelProps) => (
...
-          size={props.titleSize || 'l'}
-          className={props.titleClassName || ''}
-        >
-          <h3>{props.title}</h3>
-        </EuiTitle>
+        <EuiFlexGroup justifyContent="flexStart" alignItems="center">
+          <EuiFlexItem grow={false}>
+            <EuiTitle
+              size={props.titleSize || 'l'}
+              className={props.titleClassName || ''}
```
ylwu-amzn, line 58: Can we set the title's default className as `preview-title` ? I see most parts use this class. We can avoid some extra work by doing this. 

Similar for subTitle class, 
```
.content-panel-subtitle {
  font-size: 12pt;
  color: #69707D;
}
```

package.json
yizheliu-amazon: adding `@elastic/charts` to "dependencies" works. no need to move to `react-vis`

public/components/ContentPanel/ContentPanel.tsx
yizheliu-amazon: Sure. will change default className to `preview-title`, and add 'content-panel-subtitle' as default className

package.json
ylwu-amzn: Cool, that's great!

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,293 @@
...
+    setIsLoading(false);
+  };
+
+  useEffect(() => {
+    async function initializeAnomaliesChart() {
+      const liveDetectors = await getLiveDetectors();
+      getLiveAnomalyData(liveDetectors);
+      setTimer(
+        setInterval(getLiveAnomalyData, MIN_IN_MILLI_SECS, liveDetectors)
+      );
```
mihirsoni, line 178: Let's add TODO to make this smart. We can stop this and save backend calls if the browser tab is not focused. i.e 
let's say I am on this page but when switch tab , all the calls we make will be waste and not really useful, can we pause it and start it again when user comes back  ? Another way we can introduce back-off mechanism.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,293 @@
...
+        style={{
+          height: '300px',
+          width: '100%',
+          opacity: 1,
+        }}
+      >
+        {isLoading ? (
+          <EuiText>
+            <EuiTextAlign textAlign="center">
+              <p>Loading anomaly data...</p>
```
mihirsoni, line 249: I think we are using loading icon. https://elastic.github.io/eui/#/display/loading

public/pages/Dashboard/Container/Dashboard.tsx
```diff
@@ -13,52 +13,42 @@
...
+import { EuiPage, EuiPageBody, EuiLoadingSpinner } from '@elastic/eui';
 import { SideBar } from '../Components/utils/SideBar';
 import { DashboardHeader } from '../Components/utils/DashboardHeader';
+import { DashboardOverview } from './DashboardOverview';
 
 export const Dashboard = () => {
   const dispatch = useDispatch();
 
-  const isLoading = useSelector((state: AppState) => state.ad.requesting);
+  const [isLoading, setIsLoading] = useState(true);
```
mihirsoni, line 30: Question: Could we be consistent , where to fetch and rely on data is it Redux or local states? I am seeing this entire place is mostly getting away with Redux and relying on local stats.

public/pages/Dashboard/utils.ts
```diff
@@ -0,0 +1,314 @@
...
+    },
+    size: 30,
+    sortField: AD_DOC_FIELDS.DATA_START_TIME,
+    from: 0,
+    sortDirection: SORT_DIRECTION.DESC,
+  };
+};
+
+export type RgbColor = [number, number, number, number?];
+export const rgbColors: RgbColor[] = [
```
mihirsoni, line 40: Question: Why do we need this ?

public/pages/DetectorResults/utils/constants.ts
```diff
@@ -13,12 +13,15 @@
...
+import {
+  SORT_DIRECTION,
+  AD_DOC_FIELDS,
+} from '../../../../server/utils/constants';
 
 export const DEFAULT_QUERY_PARAMS = {
   from: 0,
   search: '',
-  size: 20,
+  size: 30,
```
mihirsoni, line 24: Question: Any specific reason on getting 30 v/s 20 ?

public/components/ContentPanel/ContentPanel.tsx
```diff
@@ -22,11 +22,16 @@ import {
...
   EuiPanel,
   EuiTitle,
+  EuiText,
+  EuiBadge,
 } from '@elastic/eui';
 
 type ContentPanelProps = {
   title: string;
   titleSize?: EuiTitleSize;
+  isLive?: boolean;
```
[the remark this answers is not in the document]
ylwu-amzn: Same feeling, here is to add some icon after the panel title, similarly we may add some link or others. How about we make it more general like titleSuffix?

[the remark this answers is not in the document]
mihirsoni: Or even better, how about we accept node to `displayTitl e`  ? Consumer can define their own way of displaying title. So it could be either String (Current behavior) or A node consumer takes over control.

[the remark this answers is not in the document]
ylwu-amzn: I'm ok for that, actually I used your solution for subtitle to just display anything passed to ContentPanel (in my case, I need to add some links), so we can get most flexibility. For our current use cases, we just add live icon or some link to title, will result in some duplicate code, but I'm ok about that.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const getAnomalyResult = async (currentDetectors: Detector[]) => {
+    const currentAnomalyResult = [] as object[];
+    const indicesSet = new Set();
+    for (let detector of currentDetectors) {
+      const resp = await dispatch(
+        getDetectorResults(detector.id, getRecentAnomalyResultQuery)
+      );
+      currentAnomalyResult.push({
+        detector: detector,
+        results: resp.data.response.results,
```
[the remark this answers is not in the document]
yizheliu-amazon: will change it.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const dispatch = useDispatch();
+
+  const [anomalyResult, setAnomalyResult] = useState([] as object[]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [indicesNumber, setIndicesNumber] = useState(0);
+
+  const timeRangeOptions = [
+    { value: '24h', text: 'Last 24 hours' },
+    { value: '7d', text: 'Last 7 days' },
+  ];
```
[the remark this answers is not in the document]
yizheliu-amazon: will do that

public/components/ContentPanel/ContentPanel.tsx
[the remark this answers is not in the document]
yizheliu-amazon: > Or even better, how about we accept node to `displayTitle` ? Consumer can define their own way of displaying title. So it could be either String (Current behavior) or A node consumer takes over control.

I think we can accept `title` as either string or a node, to make it backwards compatible.

@mihirsoni and @ylwu-amzn feel free to let me know if that sounds good to you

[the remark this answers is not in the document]
ohltyler: I have a similar problem as well with the title here: https://aws-uxdr.invisionapp.com/share/MZVZS92PYSX#/screens/405140830. I have a soln right now which is similar to Yizhe's by adding a field in the ContentPanelProps. Let me know what you guys decide on.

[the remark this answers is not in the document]
ylwu-amzn: > > Or even better, how about we accept node to `displayTitle` ? Consumer can define their own way of displaying title. So it could be either String (Current behavior) or A node consumer takes over control.
> 
> I think we can accept `title` as either string or a node, to make it backwards compatible.
> 
> @mihirsoni and @ylwu-amzn feel free to let me know if that sounds good to you

Sounds good for me.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+      );
+      currentAnomalyResult.push({
+        detector: detector,
+        results: resp.data.response.results,
+      });
+      indicesSet.add(detector.indices.toString());
+    }
+    setAnomalyResult(currentAnomalyResult);
+    setIndicesNumber(indicesSet.size);
+    setIsLoading(false);
```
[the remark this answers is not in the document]
yizheliu-amazon: The problem is the that on the Dashboard page, if there are 2 or more components using same redux, I will encounter the following issues:

Issue 1: infinite refreshing. Dashboard.tsx and AnomaliesLiveChart.tsx is in the same Dashboard page. If both Dashboard.tsx and AnomaliesLiveChart.tsx uses `const isLoading = useSelector((state: AppState) => state.ad.requesting);` and it cause infinite refreshing of the page because both components use that `ad` redux to get data, but `ad.requesting` keeps changing, and Dashboard page keeps refreshing.

Issue 2: always refresh both components AnomaliesLiveChart.tsx and AnomaliesDistribution.tsx, while only AnomaliesLiveChart needs refreshing every 1min, if both of them use  `const isLoading = useSelector((state: AppState) => state.anomalyResults.requesting);`. My understanding is that this is caused by auto refresh of AnomaliesLiveChart, and `state.anomalyResults.requesting` is shared by those 2 components. Once `requesting` changes to true by AnomaliesLiveChart, the UI of AnomaliesDistribution is changed as well, but AnomaliesDistribution does not make API call. 

Those are the reasons I avoid using the global redux. And code here is my workaround. Please let me know your thoughts on how to use such redux appropriately.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+        style={{
+          height: '500px',
+          width: '70%',
+          opacity: 1,
+        }}
+      >
+        <EuiFlexGroup style={{ padding: '10px' }}>
+          <EuiFlexItem>
+            <h5>Index</h5>
+            <h4>{isLoading ? '-' : indicesNumber.toString()}</h4>
```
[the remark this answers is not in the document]
yizheliu-amazon: Cool. thanks! Will change that.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,293 @@
...
+    if (numDetectors < 1) {
+      return detectorArray;
+    }
+    const liveDetectorIds: string[] = get(searchResults, 'hits.hits', []).map(
+      (result: any) => result._source.detector_id
+    );
+
+    for (let detectorId of liveDetectorIds) {
+      const getDetectorResp = await dispatch(getDetector(detectorId));
+      detectorArray.push(getDetectorResp.data.response);
```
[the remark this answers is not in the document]
yizheliu-amazon: No. we only allow at most 10 detectors for this component

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,293 @@
...
+  const getLiveAnomalyData = async (currentDetectors: Detector[]) => {
+    // clear previous result
+    const currentLiveAnomalyData = [] as object[];
+    for (let detector of currentDetectors) {
+      const resp = await dispatch(
+        getDetectorResults(detector.id, getRecentAnomalyResultQuery)
+      );
+      currentLiveAnomalyData.push({
+        detector: detector,
+        results: resp.data.response.results,
```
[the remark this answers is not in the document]
yizheliu-amazon: will do.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -0,0 +1,293 @@
...
+        ],
+      },
+    },
+    sort: {
+      [AD_DOC_FIELDS.DATA_START_TIME]: SORT_DIRECTION.DESC,
+    },
+    collapse: {
+      field: AD_DOC_FIELDS.DETECTOR_ID,
+    },
+  };
```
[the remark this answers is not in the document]
yizheliu-amazon: will move it to utils/helper file.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
yizheliu-amazon: hmmm, that is good point. For now we don;t have requirement to support multi tabs, but this definitely should be considered in the future

yizheliu-amazon: I will add TODO for it.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
yizheliu-amazon: Thanks. I will add it.

public/pages/Dashboard/utils.ts
yizheliu-amazon: The sunburst chart of EUI I am using requires us to manually provide color palette: https://elastic.github.io/elastic-charts/?path=/story/sunburst--sunburst-with-two-layers

public/pages/DetectorResults/utils/constants.ts
yizheliu-amazon: I think I changed it by mistake. I need 30 anomaly results as the Live Chart needs results within last 30 mins, so we need at most 30 results given the smallest interval - 1 mins. Will revert it back to 20.

public/pages/Dashboard/Container/Dashboard.tsx
yizheliu-amazon: Explained as above. Please share your thoughts on how to use same redux across multiple component on same page.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  useEffect(() => {
+    getAnomalyResult(props.selectedDetectors);
+  }, [timeRange, props.selectedDetectors]);
+
+  return (
+    <ContentPanel
+      title="Anomalies by index and detector"
+      titleSize="s"
+      subTitle="The inner circle shows the anomaly distribution by your indices. The outer circle shows the anomaly distribution by your detector"
```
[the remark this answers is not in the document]
yizheliu-amazon: Sure. will check with tech writer.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+          id="timeRangeSelect"
+          options={timeRangeOptions}
+          value={timeRange}
+          onChange={handleOnChange}
+        />
+      }
+    >
+      <div
+        style={{
+          height: '500px',
```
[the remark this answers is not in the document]
yizheliu-amazon: sure. I can try with Flex.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,163 @@
...
+        anomalyResult.detector.indices.toString();
+        visualizedResult.push({
+          [AD_DOC_FIELDS.DETECTOR_NAME]: anomalyResult.detector.name,
+          [AD_DOC_FIELDS.INDICES]: anomalyResult.detector.indices.toString(),
+          count: anomalyResult.results.length,
+        });
+      }
+    );
+    return visualizedResult;
+  };
```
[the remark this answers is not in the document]
yizheliu-amazon: will do

public/components/ContentPanel/ContentPanel.tsx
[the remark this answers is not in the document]
mihirsoni: > > Or even better, how about we accept node to `displayTitle` ? Consumer can define their own way of displaying title. So it could be either String (Current behavior) or A node consumer takes over control.
> 
> I think we can accept `title` as either string or a node, to make it backwards compatible.
> 
> @mihirsoni and @ylwu-amzn feel free to let me know if that sounds good to you

Sounds good to me as well.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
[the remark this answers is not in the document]
mihirsoni: I'll need to rethink this but to unblock, can we name isLoading to appropriate to relate better such as `setAnomalyResultsLoading`

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -0,0 +1,369 @@
...
+    },
+    size: 30,
+    sortField: AD_DOC_FIELDS.DATA_START_TIME,
+    from: 0,
+    sortDirection: SORT_DIRECTION.DESC,
+  };
+};
+
+export type RgbColor = [number, number, number, number?];
+export const rgbColors: RgbColor[] = [
```
mihirsoni, line 47: Looking at their utils there is optimal way to generate this, could you take a look at it ?

mihirsoni: https://github.com/elastic/elastic-charts/blob/5a9ec7d34a11876766c53c10fea00b4305567189/stories/utils/utils.ts

yizheliu-amazon: > https://github.com/elastic/elastic-charts/blob/5a9ec7d34a11876766c53c10fea00b4305567189/stories/utils/utils.ts

I am not able to import methods from the utils. That is why I implemented such code by referring to code from elastic/charts [code](https://github.com/elastic/elastic-charts/blob/master/src/chart_types/partition_chart/layout/utils/calcs.ts#L24), and color combinations from [here](https://github.com/elastic/elastic-charts/blob/master/src/mocks/hierarchical/palettes.ts#L22). It would be better if there is anyway I can directly import their utils. Please let me know if you have idea on how to import

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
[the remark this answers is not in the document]
yizheliu-amazon: > I'll need to rethink this but to unblock, can we name isLoading to appropriate to relate better such as `setAnomalyResultsLoading`
Sure. Will rename it.
> Few question:
> 
> * Shouldn't both charts refresh together always ? (i.e let's if we want to make some live charts, it has to be in sync and on same time. IMO.
I checked with UX, she confirmed that only Live chart should be refreshed every min, while this distribution sunburst chart is static unless user chooses new detectors

public/pages/Dashboard/utils.ts
yizheliu-amazon: As discussed offline, since there is no good way for us to leverage utils from elastic chart, our own implementation is the only solution for now.

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const visibleIndices = get(elasticsearchState, 'indices', []) as CatIndex[];
+  const visibleAliases = get(elasticsearchState, 'aliases', []) as IndexAlias[];
+  const allIndices = [
+    ...visibleIndices
+      .filter(value => isSystemIndices(value.index))
+      .map(value => value.index),
+    ...visibleAliases
+      .filter(value => isSystemIndices(value.index))
+      .map(value => value.index),
+  ];
```
ohltyler, line 135: use getVisibleOptions() helper fn here? see https://github.com/ohltyler/anomaly-detection-kibana-plugin/blob/updateDetectorListPage/public/pages/DetectorsList/List/List.tsx#L101

yizheliu-amazon: it will return list of index options for combo box, which is not needed for filtering case here. I do use getVisibleOptions() in line 237 for combo box

ohltyler: I see now, I did it a little different where I just ignore the filter altogether if the page state currently has all indices selected (in your case, allIndicesSelected set to true). This could maybe be a little more space efficient and eliminate some lines of code since you wouldn't need allIndices, but a small thing I guess.

public/pages/Dashboard/utils/constants.ts
```diff
@@ -0,0 +1,106 @@
...
+      gte: 'now-30m',
+    },
+  },
+  size: 30,
+  sortField: AD_DOC_FIELDS.DATA_START_TIME,
+  from: 0,
+  sortDirection: SORT_DIRECTION.DESC,
+};
+
+export const GET_ALL_RECENT_ANOMALOUS_DETECORS_QUERY = {
```
ohltyler, line 55: Small typo in DETECTORS here

yizheliu-amazon: good catch. Thanks

public/pages/Dashboard/Container/DashboardOverview.tsx
yizheliu-amazon: I see your. That is a good point. Will change to check if allIndicesSelected instead.

public/utils/constants.ts
```diff
@@ -38,3 +38,15 @@ export const APP_PATH = {
...
+
+export const MAX_DETECTORS = 1000;
+
+export const MAX_ANOMALIES = 10000;
+
+export const DETECTOR_STATE = Object.freeze({
+  DISABLED: 'Disabled',
+  INIT: 'Initializing',
+  RUNNING: 'Running',
+});
```
ohltyler, line 52: We should be able to just use a TypeScript enum here. From what I can tell Object.freeze() is a JavaScript workaround to create similar behavior to enums. See https://github.com/ohltyler/anomaly-detection-kibana-plugin/blob/updateDetectorListPage/public/pages/utils/constants.ts#L18

yizheliu-amazon: sure. will use the same type as yours.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -0,0 +1,242 @@
...
+    const resultDetectors = getFinalDetectors(
+      finalAnomalyResult,
+      props.selectedDetectors
+    );
+    setIndicesNumber(getFinalIndice(resultDetectors).size);
+    setFinalDetectors(resultDetectors);
+    setAnomalyResultsLoading(false);
+  };
+
+  const getFinalIndice = (detectorList: DetectorListItem[]) => {
```
ohltyler, line 84: Maybe change to getFinalIndices()?

public/pages/Dashboard/Components/AnomalousDetectorsList.tsx
```diff
@@ -0,0 +1,95 @@
...
+    sort: {
+      direction: sortDirection,
+      field: fieldForSort,
+    },
+  };
+  const pagination = {
+    pageIndex: indexOfPage,
+    pageSize: sizeOfPage,
+    totalItemCount: Math.min(MAX_DETECTORS, props.selectedDetectors.length),
+    pageSizeOptions: [10, 20, 50],
```
ohltyler, line 47: Nitpick: change to [5, 10, 20, 50] for consistency w/ the detector list page?

yizheliu-amazon: No preference from me. Will confirm with UX for this.

yizheliu-amazon: Confirmed with UX, will change to [5, 10, 20, 50]

server/routes/ad.ts
```diff
@@ -398,51 +398,81 @@ const getAnomalyResults = async (
...
     let sort = {};
     const sortQuery = sortQueryMap[sortField];
     if (sortQuery) {
       sort = sortQuery;
     }
+
+    let rangeObj = range;
+
+    if (range !== undefined && typeof range === 'string') {
+      rangeObj = JSON.parse(range);
```
mihirsoni, line 435: Are we sure this will be always valid JSON string ? If not then wrap this inside `try...catch` this will fail JSON.parse("hello")

public/utils/constants.ts
mihirsoni: +1

public/components/ContentPanel/index.scss
```diff
@@ -0,0 +1,10 @@
+.content-panel-title {
```
ylwu-amzn, line 1: Copyright header?

public/pages/Dashboard/index.scss
```diff
@@ -0,0 +1,31 @@
+.live-anomaly-results-subtile {
```
ylwu-amzn, line 1: Copyright header?

public/components/ContentPanel/index.scss
yizheliu-amazon: will add it.

server/routes/ad.ts
yizheliu-amazon: It is already inside `try catch`, and it does log the error in case of any exception and then
```
return { ok: false, error: err.message }
```
The question is: given `range` is not undefined, should we fail the request, if `range` is invalid string? If yes, then we may keep current implementation; if no, aka we just ignore invalid `range` string. I would probably go with `Failing the request` for now, because `range` is specified, but invalid input, and we should fail it to let caller know the reason, instead of swallowing issue and returning result caller doesn't expect.

public/pages/Dashboard/index.scss
yizheliu-amazon: will add it.

server/utils/helpers.ts
```diff
@@ -20,8 +20,11 @@ import {
   mapKeys,
   mapValues,
   snakeCase,
+  get,
```
ylwu-amzn, line 23: Seems `get` is not used in this file. Remove it?

yizheliu-amazon: Done

PR #18, opened 2020-04-09
add search monitor api;change detector enable/disalbe time to number

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
1. Add alerting search monitor API.
2. Change detector enabled/disabled time to number


server/routes/alerting.ts
```diff
@@ -0,0 +1,89 @@
...
+      query: {
+        nested: {
+          path: 'monitor.inputs',
+          query: {
+            bool: {
+              must: [
+                {
+                  term: {
+                    'monitor.inputs.search.indices.keyword': {
+                      value: '.opendistro-anomaly-results*',
```
mihirsoni, line 45: Why search value is index name ?

server/routes/alerting.ts
```diff
@@ -0,0 +1,89 @@
...
+      { body: requestBody }
+    );
+    const totalMonitors = get(response, 'hits.total.value', 0);
+    const allMonitors = get(response, 'hits.hits', []).reduce(
+      (acc: any, monitor: any) => ({
+        ...acc,
+        [monitor._id]: {
+          id: monitor._id,
+          name: get(monitor, '_source.name', ''),
+          enabled: get(monitor, '_source.enabled', ''),
```
mihirsoni, line 67: default should be a Boolean value as per the Model.

server/routes/alerting.ts
```diff
@@ -0,0 +1,89 @@
...
+          enabledTime: get(monitor, '_source.enabled_time', ''),
+          schedule: get(monitor, '_source.schedule', ''),
```
mihirsoni, line 69: Default value should be `period {}`

ylwu-amzn: Thanks, will change it

server/routes/alerting.ts
ylwu-amzn: For AD monitors, we will use AD result index and we can search AD monitors with this.

ylwu-amzn: One sample for monitor inputs
```
...
"inputs" : [
            {
              "search" : {
                "indices" : [
                  ".opendistro-anomaly-results*"
                ],
                "query" : {
                  "size" : 1,
                  "query" : {
                    "bool" : {
                      "filter" : [
                        {
                          "range" : {
                            "data_end_time" : {
                              "from" : "{{period_end}}||-2m",
                              "to" : "{{period_end}}",
                              "include_lower" : true,
                              "include_upper" : true,
                              "boost" : 1.0
                            }
                          }
                        },
                        {
                          "term" : {
                            "detector_id" : {
                              "value" : "LNULDnEBdsd5fFw6Wr6N",
                              "boost" : 1.0
                            }
                          }
                        }
                      ],
                      "adjust_pure_negative" : true,
                      "boost" : 1.0
                    }
                  },
                  "sort" : [
                    {
                      "anomaly_grade" : {
                        "order" : "desc"
                      }
                    },
                    {
                      "confidence" : {
                        "order" : "desc"
                      }
                    }
                  ],
                  "aggregations" : {
                    "max_anomaly_grade" : {
                      "max" : {
                        "field" : "anomaly_grade"
                      }
                    }
                  }
                }
              }
            }
          ]
...
```

server/routes/alerting.ts
ylwu-amzn: Will change to false.

server/routes/alerting.ts
ylwu-amzn: Has removed unnecessary default values, just keep monitor enabled default value as AD only needs this currently.

PR #19, opened 2020-04-10
home page and side nav bar hook

ylwu-amzn:
*Issue #, if available:*

* add side nav bar to home page
* control show/hide side bar through AD app hook, sample code to use it
```
...
import { useHideSideNavBar } from '../../DetectorDetail/hooks/useHideSideNavBar';
...

export function EditFeatures(props: EditFeaturesProps) {
  // the first boolean of useHideSideNavBar is true, means hide side nav bar on this page
 // the second boolean of useHideSideNavBar is false means show side nav bar after exit this page
  useHideSideNavBar(true, false);
  ...
}
```

### DIFF
--- a/public/pages/main/Main.tsx
+++ b/public/pages/main/Main.tsx
@@ -15,50 +15,106 @@
 
 import { Switch, Route, RouteComponentProps, Redirect } from 'react-router-dom';
 import React from 'react';
+import { AppState } from '../../redux/reducers';
 import { CreateDetector } from '../createDetector';
-
 import { DetectorList } from '../DetectorsList';
 import { ListRouterParams } from '../DetectorsList/List/List';
-import { PreviewDetector } from '../PreviewDetector/containers/PreviewDetector';
-import { Dashboard } from '../Dashboard/Container/Dashboard';
-import { APP_PATH } from '../../utils/constants';
+// @ts-ignore
+import { EuiSideNav, EuiPage, EuiPageBody, EuiPageSideBar } from '@elastic/eui';
+import { useSelector } from 'react-redux';
 
-interface MainProps {}
+enum Navigation {
+  AnomalyDetection = 'Anomaly detection',
+  Dashboard = 'Dashboard',
+  Detectors = 'Detectors',
+}
 
-export function Main(mainProps: MainProps) {
+enum Pathname {
+  Dashboard = '/dashboard',
+  Detectors = '/detectors',
+}
+
+interface MainProps extends RouteComponentProps {}
+
+export function Main(props: MainProps) {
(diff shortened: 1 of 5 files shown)


public/pages/main/Main.tsx
```diff
@@ -15,50 +15,106 @@
...
+  AnomalyDetection = 'Anomaly detection',
+  Dashboard = 'Dashboard',
+  Detectors = 'Detectors',
+}
 
-interface MainProps {}
+enum Pathname {
+  Dashboard = '/dashboard',
+  Detectors = '/detectors',
+}
```
yizheliu-amazon, line 35: nit: move them to public/utils/constants.ts

public/pages/main/Main.tsx
```diff
@@ -15,50 +15,106 @@
...
+    <EuiPage>
+      <EuiPageSideBar style={{ minWidth: 150 }} hidden={hideSideNavBar}>
+        <EuiSideNav style={{ width: 150 }} items={sideNav} />
+      </EuiPageSideBar>
+      <EuiPageBody>
+        <Switch>
+          <Route
+            path="/dashboard"
+            render={(props: RouteComponentProps) => (
+              // place holder for DashboardOverview, please replace with your page
```
yizheliu-amazon, line 75: thanks!

public/pages/main/Main.tsx
```diff
@@ -15,50 +15,106 @@
...
+  Dashboard = '/dashboard',
+  Detectors = '/detectors',
+}
+
+interface MainProps extends RouteComponentProps {}
+
+export function Main(props: MainProps) {
+  const hideSideNavBar = useSelector(
+    (state: AppState) => state.adApp.hideSideNavBar
+  );
```
mihirsoni, line 42: @ylwu-amzn  Do you think we would need this state globally ? I think it can be just use local state v/s global state in redux. Do you need this value of state outside of this component ? If not let's just `useState` hooks from react and remove all Redux layer if not really needed.

ylwu-amzn: Thanks, we need to use this state for multiple components/pages. For the home page, we should show side nav bar, but for detector detail, AD result page, edit detector/feature pages, we should hide the side nav bar. So I think we need a global state in redux.

public/pages/main/Main.tsx
ylwu-amzn: sure, will change it.

public/pages/main/Main.tsx
ylwu-amzn: Once this code merged, you can change to 
```
render={(props: RouteComponentProps) => <DashboardOverview />}
```

Issue #29, opened 2020-04-17
Width of Dashboard is fixed so that content can get hidden if browser window shrinks


yizheliu-amazon:
Width of Dashboard is somehow fixed so that contents on right side can get hidden if browser window shrinks, like `Create Detector` button, part of `Live chart` and `Detector and features` table

![Screen Shot 2020-04-16 at 6 13 00 PM](https://user-images.githubusercontent.com/59710443/79521915-2f05de80-800f-11ea-9ea9-92668e5a7bc1.png)

PR #30, opened 2020-04-17
Fix issue that width of Dashboard is somehow fixed. Issue: #29

yizheliu-amazon:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/29
*Description of changes:*
Fix issue that width of Dashboard is somehow fixed, which causes right-side components will get hidden or partially hidden.

After fix, even if the window shrinks, the button is still visible. 
Screen Shot 2020-04-16 at 6 26 11 PM

### DIFF
--- a/public/pages/Dashboard/index.scss
+++ b/public/pages/Dashboard/index.scss
@@ -24,13 +24,10 @@
 }
 
 .anomaly-distribution-subtitle {
-  height: 28px;
-  width: 505px;
   color: #879196;
   font-family: 'Helvetica Neue';
   font-size: 12px;
   letter-spacing: 0;
-  line-height: 14px;
 }
 
 .live-badge {
--- a/public/pages/Dashboard/index.scss
+++ b/public/pages/Dashboard/index.scss
@@ -30,13 +30,6 @@
   letter-spacing: 0;
 }
 
-.live-badge {
-  height: 20px;
-  width: 46px;
-  border-radius: 2px;
-  background-color: #db1374;
-}
-
 .anomalies-distribution-sunburst {
   height: 443px;
   width: 441px;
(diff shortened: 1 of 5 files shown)


public/pages/main/Main.tsx
```diff
@@ -70,7 +70,7 @@ export function Main(props: MainProps) {
       <EuiPageSideBar style={{ minWidth: 150 }} hidden={hideSideNavBar}>
         <EuiSideNav style={{ width: 150 }} items={sideNav} />
       </EuiPageSideBar>
-      <EuiPageBody>
+      <EuiPageBody restrictWidth={'70%'}>
```
ylwu-amzn, line 73: How about we remove this restriction to get wider page, especially for detector list

yizheliu-amazon: Good point. It turns out that the min width of bottom 2 components `sunburst chart` + `detector feature list` is around 800px. And min width of Side Nav Bar is 150px. The smallest width which has no hidden contents is around 1000px.

Screenshot shows the above smallest width on mac 13in screen. 

Screen Shot 2020-04-16 at 9 16 39 PM

yizheliu-amazon: Is that smallest width acceptable?

ylwu-amzn: I think that's ok considering current popular resolution is above 1000px

public/pages/Dashboard/index.scss
```diff
@@ -24,13 +24,10 @@
...
-  height: 28px;
-  width: 505px;
   color: #879196;
   font-family: 'Helvetica Neue';
   font-size: 12px;
   letter-spacing: 0;
-  line-height: 14px;
 }
 
 .live-badge {
```
ylwu-amzn, line 33: Do we still need this ?

yizheliu-amazon: no. Good catch. will remove it.

PR #42, opened 2020-04-23
Add unit test workflow

ohltyler:
*Issue #, if available:*

*Description of changes:*

- Updates test snapshots
- Adds a GitHub action to run all unit tests on multiple OS

**NOTE:** The workflow clones the private ODFE Kibana repository, so a secret token must be passed to the GitHub runner for authentication. Because of this, the workflow cannot be ran when a pull request is made (see https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets). This workflow is currently configured to run every time a push is made to the development branch - it can be treated as a final check after a PR is approved & merged.

### DIFF
--- a/.github/workflows/unit-tests-workflow.yml
+++ b/.github/workflows/unit-tests-workflow.yml
@@ -1,0 +1,48 @@
+name: Unit tests workflow
+on:
+  push:
+    branches:
+      - development
+
+jobs:
+  tests:
+    name: Run unit tests
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout Kibana
+        uses: actions/checkout@v2
+        with:
+          repository: opendistro-for-elasticsearch/kibana-oss
+          ref: 7.4.2
+          token: ${{ secrets.KIBANA_OSS_ACCESS }}
+          path: kibana
+      - name: Get node and yarn versions
+        id: versions_step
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests-workflow.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          path: kibana/plugins/anomaly-detection-kibana-plugin
+      - name: Bootstrap the plugin
+        run: |
+          cd kibana/plugins/anomaly-detection-kibana-plugin
+          yarn kbn bootstrap
+      - name: Run tests
+        run: |
+          cd kibana/plugins/anomaly-detection-kibana-plugin
+          yarn run test:jest
```
ylwu-amzn, line 48: This will run when PR published and fail if any snapshot tests fail?

ohltyler: It will only run (and only can be ran) when pushed to the development branch, used as a final check (see description above). It will fail if any snapshot tests fail, yes.

yizheliu-amazon:
The snapshot changes seem to be same as #44 . Should we discard this one? Btw, good work for setting up the actions!

ohltyler:
> The snapshot changes seem to be same as #44 . Should we discard this one and create another PR with workflow setup yml file only?
> 
> Btw, good work for setting up the actions!

Thanks - and yes, will rebase when the update snapshots PR is merged

Issue #37, opened 2020-04-21
[Dashboard] Remove “10 detectors with the most recent anomaly occurrence” from dashboard live chart if no data.


yizheliu-amazon:
Remove “10 detectors with the most recent anomaly occurrence” from dashboard live chart if no data.

yizheliu-amazon:
new comment:

If no data found, we show “no data to display”;

if we found data, but all zero, we should show the live chart like detector live anomaly chart. And show some words like “No anomaly occurrences are found in last 30 minutes”. 

Dynamically show `number` of most recent anomalous detectors even less than 10 detectors

Issue #36, opened 2020-04-21
[Dashboard] Dashboard always show loading status if there is no AD ever created


yizheliu-amazon:
If there is no detector created before, user of 1st time entering Dashboard will see loading status instead of empty Dashboard page. The root cause is that, in that case, checking if there is detector fails due to exception: [index_not_found_exception] no such index [.opendistro-anomaly-detectors], and we fail to handle such exception gracefully. 

Fix: show empty Dashboard page in case of index_not_found_exception

Issue #26, opened 2020-04-15
Add support for "View Full Screen" for Dashboard Live chart


yizheliu-amazon:
Add support for "View Full Screen" for Dashboard Live chart

mihirsoni:
Should we create different issue for managing performance for the live chart ?

yizheliu-amazon:
> Should we create different issue for managing performance for the live chart ?

Can you elaborate what is the performance issue for live chart?

PR #45, opened 2020-04-24
Address couples of feedback. Issue: #26, #36, #37, #38

yizheliu-amazon:
*Issue #, if available:*
Address couples of feedback. Issue: #26, #36, #37, #38
*Description of changes:*
1. change the detector link on dashboard detector list table should redirect to AD result page
1. add Fullscreen of dashboard live chart
1. Fix issue that: for a new domain which has no AD plugin running, dashboard will always show loading status as we can’t get detectors (detector index not found). We need to show empty dashboard page
1. Decrease line height for subtitle of Distribution Sunburst Chart on Dashboard
1. Increase margin Live Anomalies title to 20px, make all the text/wording is aligned/consistent with other components
1. Filter boxes can use full-width, margin between filter boxes is too wide(decrease to 8px), we can make filter boxes take the all whole row. The goal is to make all boxes long enough to avoid stacked multi-selections as much as possible
1. Use dynamic height for Detector&Feature list, its height is not necessary to be same as Distribution Sunburst chart.
1. Decrease size of Distribution Sunburst chart to 400px/360px
1. Change style of subtitle of Live Anomaly chart.
1. Keep XL loading chart only.
1.  Couples of style changes required by UX designer. 
1. Change anomaly live chart in cases of 1) no anomaly result; 2) there exists anomaly data; 3) only anomaly grade 0 data like below

Screen Shot 2020-04-23 at 9 45 19 PM
Screen Shot 2020-04-23 at 10 42 38 PM
Screen Shot 2020-04-23 at 11 09 48 PM

### DIFF
--- a/public/components/ContentPanel/ContentPanel.tsx
+++ b/public/components/ContentPanel/ContentPanel.tsx
@@ -38,14 +38,16 @@ type ContentPanelProps = {
   titleContainerStyles?: React.CSSProperties;
   actions?: React.ReactNode | React.ReactNode[];
   children: React.ReactNode | React.ReactNode[];
+  contentPanelClassName?: string;
 };
 
 const ContentPanel = (props: ContentPanelProps) => (
   <EuiPanel
     style={{ paddingLeft: '0px', paddingRight: '0px', ...props.panelStyles }}
+    className={props.contentPanelClassName}
   >
     <EuiFlexGroup
-      style={{ padding: '0px 10px', ...props.titleContainerStyles }}
+      style={{ padding: '0px 20px', ...props.titleContainerStyles }}
       justifyContent="spaceBetween"
       alignItems="center"
     >
@@ -105,7 +107,7 @@ const ContentPanel = (props: ContentPanelProps) => (
         className={props.horizontalRuleClassName}
       />
     )}
-    <div style={{ padding: '0px 10px', ...props.bodyStyles }}>
+    <div style={{ padding: '0px 20px', ...props.bodyStyles }}>
       {props.children}
     </div>
   </EuiPanel>
--- a/public/components/ContentPanel/ContentPanel.tsx
+++ b/public/components/ContentPanel/ContentPanel.tsx
@@ -14,8 +14,9 @@
  */
 
 import React from 'react';
-//@ts-ignore
+
 import {
+  //@ts-ignore
   EuiTitleSize,
   EuiFlexGroup,
   EuiFlexItem,
@@ -43,7 +44,10 @@ type ContentPanelProps = {
 
 const ContentPanel = (props: ContentPanelProps) => (
   <EuiPanel
-    style={{ paddingLeft: '0px', paddingRight: '0px', ...props.panelStyles }}
+    style={{
+      padding: '20px 0px',
+      ...props.panelStyles,
+    }}
     className={props.contentPanelClassName}
   >
     <EuiFlexGroup
@@ -55,7 +59,7 @@ const ContentPanel = (props: ContentPanelProps) => (
         {typeof props.title === 'string' ? (
           <EuiTitle
             size={props.titleSize || 'l'}
-            className={props.titleClassName || 'content-panel-title'}
+            className={props.titleClassName || ''}
           >
             <h3>{props.title}</h3>
           </EuiTitle>
@@ -107,7 +111,14 @@ const ContentPanel = (props: ContentPanelProps) => (
         className={props.horizontalRuleClassName}
       />
     )}
-    <div style={{ padding: '0px 20px', ...props.bodyStyles }}>
+    <div
+      style={{
+        paddingTop: '12px',
+        paddingLeft: '20px',
+        paddingRight: '20px',
+        ...props.bodyStyles,
+      }}
+    >
       {props.children}
     </div>
   </EuiPanel>
--- a/public/components/ContentPanel/index.scss
+++ b/public/components/ContentPanel/index.scss
@@ -14,12 +14,11 @@
  */
 
 .content-panel-title {
-  color: #3f3f3f;
+  color: #000;
 }
 
 .content-panel-subTitle {
   color: #879196;
-  font-family: 'Helvetica Neue';
   font-size: 12px;
   letter-spacing: 0;
 }
--- a/public/components/ContentPanel/index.scss
+++ b/public/components/ContentPanel/index.scss
@@ -13,10 +13,6 @@
  * permissions and limitations under the License.
  */
 
-.content-panel-title {
-  color: #000;
-}
-
 .content-panel-subTitle {
   color: #879196;
   font-size: 12px;
--- a/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
@@ -34,9 +34,9 @@ import { useDispatch } from 'react-redux';
 import { Datum } from '@elastic/charts/dist/utils/commons';
 import React from 'react';
 import { TIME_RANGE_OPTIONS } from '../../Dashboard/utils/constants';
-import { get } from 'lodash';
+import { get, isEmpty } from 'lodash';
 import { searchES } from '../../../redux/reducers/elasticsearch';
-import { MAX_DETECTORS } from '../../../utils/constants';
+import { MAX_DETECTORS, MAX_ANOMALIES } from '../../../utils/constants';
 import { AD_DOC_FIELDS } from '../../../../server/utils/constants';
 export interface AnomaliesDistributionChartProps {
   allDetectorsSelected: boolean;
@@ -64,17 +64,23 @@ export const AnomaliesDistributionChart = (
   const [timeRange, setTimeRange] = useState(TIME_RANGE_OPTIONS[0].value);
 
   const getAnomalyResult = async (currentDetectors: DetectorListItem[]) => {
-    const finalAnomalyResult = await getLatestAnomalyResultsForDetectorsByTimeRange(
+    const latestAnomalyResult = await getLatestAnomalyResultsForDetectorsByTimeRange(
       searchES,
       props.selectedDetectors,
       timeRange,
-      MAX_DETECTORS,
-      dispatch
+      dispatch,
+      0,
+      MAX_ANOMALIES,
+      MAX_DETECTORS
     );
-    setAnomalyResults(finalAnomalyResult);
+
+    const nonZeroAnomalyResult = latestAnomalyResult.filter(
+      anomalyData => get(anomalyData, AD_DOC_FIELDS.ANOMALY_GRADE, 0) > 0
+    );
+    setAnomalyResults(nonZeroAnomalyResult);
 
     const resultDetectors = getFinalDetectors(
-      finalAnomalyResult,
+      nonZeroAnomalyResult,
       props.selectedDetectors
     );
     setIndicesNumber(getFinalIndices(resultDetectors).size);
@@ -123,9 +129,8 @@ export const AnomaliesDistributionChart = (
         <EuiFlexItem>
           <EuiText className={'anomaly-distribution-subtitle'}>
             <p>
-              {
-                'The inner circle shows the anomaly distribution by your indices. The outer circle shows the anomaly distribution by your detector'
-              }
+              {'The inner circle shows the anomaly distribution by your indices. ' +
+                'The outer circle shows the anomaly distribution by your detectors.'}
             </p>
           </EuiText>
         </EuiFlexItem>
@@ -141,13 +146,14 @@ export const AnomaliesDistributionChart = (
         />
       }
     >
-      <EuiFlexGroup style={{ padding: '10px' }}>
+      <EuiFlexGroup style={{ marginTop: '0px' }}>
         <EuiFlexItem>
           <EuiStat
             description={'Indices with anomalies'}
             title={indicesNumber}
             isLoading={anomalyResultsLoading}
             titleSize="s"
+            style={{ color: '#000' }}
           />
         </EuiFlexItem>
         <EuiFlexItem>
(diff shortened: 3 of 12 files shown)


public/components/ContentPanel/index.scss
```diff
@@ -14,12 +14,11 @@
  */
 
 .content-panel-title {
-  color: #3f3f3f;
+  color: #000;
```
ylwu-amzn, line 17: How about remove this color, we can use the default color to keep consistent.

public/components/ContentPanel/ContentPanel.tsx
```diff
@@ -38,14 +38,16 @@ type ContentPanelProps = {
...
 };
 
 const ContentPanel = (props: ContentPanelProps) => (
   <EuiPanel
     style={{ paddingLeft: '0px', paddingRight: '0px', ...props.panelStyles }}
+    className={props.contentPanelClassName}
   >
     <EuiFlexGroup
-      style={{ padding: '0px 10px', ...props.titleContainerStyles }}
+      style={{ padding: '0px 20px', ...props.titleContainerStyles }}
```
ylwu-amzn, line 50: How about we change this padding as 0
```
style={{ padding: '0px', ...props.titleContainerStyles }}
```

And set the padding of `EuiPanel` as `20px`, so we can align title and content body.
```
<EuiPanel
    style={{ padding: '20px', ...props.panelStyles }}
    className={props.contentPanelClassName}
  >
```

public/components/ContentPanel/ContentPanel.tsx
```diff
@@ -105,7 +107,7 @@ const ContentPanel = (props: ContentPanelProps) => (
         className={props.horizontalRuleClassName}
       />
     )}
-    <div style={{ padding: '0px 10px', ...props.bodyStyles }}>
+    <div style={{ padding: '0px 20px', ...props.bodyStyles }}>
```
ylwu-amzn, line 110: Why increase left and right padding? Shouldn't we increase the top and down padding?

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -156,85 +162,82 @@ export const AnomaliesDistributionChart = (
             title={finalDetectors.length}
             isLoading={anomalyResultsLoading}
             titleSize="s"
+            style={{ color: '#000' }}
```
ylwu-amzn, line 165: Why not use default color?

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -137,132 +161,203 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
...
         </EuiTitle>
       }
       subTitle={
         <EuiFlexItem>
           <EuiText className={'live-anomaly-results-subtile'}>
             <p>
-              {'Live anomaly results across detectors for the last 30 minutes'}
+              {'Live anomaly results across detectors for the last 30 minutes. ' +
+                'The results refresh every 1 minute. ' +
+                'For each detector, if an anomaly occurence is detected at the end of the detector interval, ' +
```
ohltyler, line 209: Typo - should be 'occurrence'

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -137,132 +161,203 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
...
+                  size="s"
+                  title={`You are viewing ${MAX_LIVE_DETECTORS} detectors with the most recent anomaly occurrences.`}
+                  style={{
+                    width: '88%', // ensure width reaches NOW annotation line
+                    marginTop: '20px',
+                    marginBottom: '20px',
+                  }}
+                >
+                  <p>
+                    10 detectors with the most recent anomalies are shown on the
```
ohltyler, line 289: Small: remove the hardcoded `10` here to be `${MAX_LIVE_DETECTORS}`

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -434,7 +435,8 @@ export const getFloorPlotTime = (plotTime: number): number => {
 export const buildGetAnomalyResultQueryByRange = (
   timeRange: string,
   from: number,
-  size: number
+  size: number,
+  threadhold: number
```
ohltyler, line 439: Small: do you mean threshold here? :)

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -477,16 +479,25 @@ export const getLatestAnomalyResultsForDetectorsByTimeRange = async (
   func: (request: any) => APIAction,
   selectedDetectors: DetectorListItem[],
   timeRange: string,
-  detectorNum = MAX_DETECTORS,
-  dispatch = useDispatch()
+  dispatch: Dispatch<any>,
+  threadhold: number,
```
ohltyler, line 483: and same thing here

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -553,27 +564,21 @@ const buildDetectorAndIdMap = (
 
 const selectLatestDetetorIds = (
```
ohltyler, line 565: Small: typo - should be `selectLatestDetectorIds`

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -553,27 +564,21 @@ const buildDetectorAndIdMap = (
...
-      )
-    ) {
-    }
-    if (latestDetectorIds.length === needeDetectorNum) {
+    const detectorId = get(anomalyData, AD_DOC_FIELDS.DETECTOR_ID, '');
+    latestDetectorIds.add(detectorId);
+    if (latestDetectorIds.size === neededDetectorNum) {
       return latestDetectorIds;
     }
   }
```
ohltyler, line 584: It looks like this unnecessarily loops through the `orderedAnomalyData` again. Can't we just slice the `uniqueIds` set to include up to `neededDetectorNum` ids? Something like `return Array.from(uniqueIds).slice(0, neededDetectorNum`.

public/components/ContentPanel/index.scss
yizheliu-amazon: sure. will remove `content-panel-title` class as it only has `color` attribute.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
yizheliu-amazon: I recalled this is required by UX to change the color, and I think that's why I use 000 for content-panel-style. I will double confirm wit UX on this.

public/components/ContentPanel/ContentPanel.tsx
yizheliu-amazon: Changing the padding of `EuiPanel` to 20px will make the `EuiHorizontalRule` looks  disconnected to the Panel. I changed the style for Live chart, and below is the screenshot. 
Screen Shot 2020-04-27 at 5 52 28 PM

This is not same as the UX mock up.

public/components/ContentPanel/ContentPanel.tsx
yizheliu-amazon: As mentioned above, since we cannot increase EuiPanel padding, we have to increase the left/right padding of `body` to 20px. 

Regarding the top/down padding, you are right, we should make them both 20px. I didn't increase them because:

1) Top padding: I set top padding in [AnomaliesLiveChart](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/45/files#diff-d0f6b17630ef8bd093c50b787e508b88R241) and [AnomaliesDistributionChart](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/45/files#diff-474a5fc2dba872f006bdc3342074dd77R149): `FlexGroup` has default padding -12px, and `EuiHorizontalRule` has default top/bottom margin 8px, changing top padding of FlexGroup to 0px can make the distance between `EuiHorizontalRule` and FlexItems in FlexGroup to 20px

2) Bottom padding: EuiPanel has default padding 16px, thus I keep it for the bottom padding.

To consolidate style changes for ContentPanel, I think I can make following changes so that all the style change can be configured in ContentPanel, and the distance between components and Panel edge or `EuiHorizontalRule` is 20px.
```
  <EuiPanel
    style={{
      paddingLeft: '0px',
      paddingRight: '0px',
      paddingTop: '20px',
      paddingBottom: '20px',
      ...props.panelStyles,
    }}
    className={props.contentPanelClassName}
  >
```

```
    <EuiFlexGroup
      style={{ padding: '0px 20px', ...props.titleContainerStyles }}
      justifyContent="spaceBetween"
      alignItems="center"
    >
```

```
    <div
      style={{
        paddingTop: '12px', // 12px here is because EuiHorizontalRule has bottom margin 8px
        paddingLeft: '20px',
        paddingRight: '20px',
        ...props.bodyStyles,
      }}
    >
      {props.children}
    </div>
```

public/pages/Dashboard/utils/utils.tsx
yizheliu-amazon: Good idea. Didn't know that the elements of a set can be iterated in insertion order. Previously I was thinking using the loop to get the 1st `neededDetectorNum` detectors with anomalies in an ordered array. Will change to such way

public/pages/Dashboard/utils/utils.tsx
yizheliu-amazon: Good catch. Thank you.

public/pages/Dashboard/utils/utils.tsx
yizheliu-amazon: Good catch. Thank you.

public/pages/Dashboard/utils/utils.tsx
yizheliu-amazon: yeah. 😂  Good catch. Thank you.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
yizheliu-amazon: Good catch. Thanks.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
yizheliu-amazon: Good catch. Thanks.

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
yizheliu-amazon: Have confirmed with UX on this. We are good to use default color.

public/components/ContentPanel/ContentPanel.tsx
ylwu-amzn: Sure, let's follow the UX mock up

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -137,132 +161,200 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
...
+                    id={'left'}
+                    title={'Anomaly grade'}
+                    position={Position.Left}
+                    domain={{ min: 0, max: 1 }}
+                  />
+                  <BarSeries
+                    // `id` for placeholder data point introduced by `prepareVisualizedAnomalies` shows as legend,
+                    // When there exists anomalies with anomaly grade > 0
+                    // we make `id` to blank string to hide the legend of placeholder data point
+                    id={!isEmpty(liveAnomalyData) ? '' : ' '}
```
ylwu-amzn, line 343: just curious: only blank string can hide legend ? Should we set a normal id rather than whitespace when the `liveAnomalyData` not empty?

public/pages/Dashboard/Container/Dashboard.tsx
```diff
@@ -29,16 +29,21 @@ export const Dashboard = () => {
...
-      })
-    );
-    setIsLoading(false);
+    try {
+      await dispatch(
+        getDetectorList({
+          from: 0,
+          size: 1,
+          search: '',
+          sortDirection: SORT_DIRECTION.DESC,
```
ylwu-amzn, line 38: Should we use ASC sort direction?  `name` is just string, is it better to show alphabetically?

yizheliu-amazon: This is just used to check if there is any detector ever created. That is why `size` is 1. If no detector, we move to EmptyDashboard page. If yes, we navigate to the real Dashboard page. using ASC or DESC doesn't matter too much here.

yizheliu-amazon: any smarter way of doing the check of existing detector is appreciated.

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
yizheliu-amazon: This is very tricky. 

If detectorName value for placeholder data is set as `null`, the `id` of `BarSeries` will be shown as legend for placeholder data.

Given empty `liveAnomalyData`, the legend of placeholder data will be hidden by changing `showLegend` to false, aka `showLegend={!isEmpty(liveAnomalyData)}`.  Thus we don't need to worry about legend of placeholder data in case of empty `liveAnomalyData`.

Given non-empty `liveAnomalyData`, the legend of placeholder data is shown along with legend of `liveAnomalyData`. Legend of placeholder data is not needed now. Changing `id` of `BarSeries` to blank string can make legend of placeholder data disappear, and normal legend of `liveAnomalyData` is still there. 

But we cannot always pass blank string `{''}`to `id`, as in the case of `null` detectorName + `id` as `{''}` , EuiChart will show `No data to dsiplay`, which UX wants to avoid. That's why  I pass `{' '}` /`{''}` to `id`

yizheliu-amazon: After second thought, I changed to 
```
[AD_DOC_FIELDS.DETECTOR_NAME]: !isEmpty(liveAnomalyData) ? '' : null,
```

and keep using
```
<BarSeries
id={'Detector Anomaly grade'}`
```
It can also work. will change to that implementation and post another commit.

public/pages/Dashboard/Container/Dashboard.tsx
ylwu-amzn: Sure, got it, that's ok. Maybe we can use `_count` in future.
https://stackoverflow.com/questions/25739888/counting-number-of-documents-using-elasticsearch
```
Probably _count is a bit faster since it doesn't have to execute a full query with ranking and result fetching and can simply return the size.
```

public/components/ContentPanel/ContentPanel.tsx
ylwu-amzn: I'm ok about it, we can tune if any other comments come in.

PR #48, opened 2020-04-28
Address detector list feedback and add feature required state

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR fixes some small feedback issues for the detector list page and adds a `Feature required` detector state. Specifically:
- Fixes the spacing issue between the detector state filter and index filter when the page is extended horizontally
- Changes the timestamps to capitalize AM/PM
- Shrinks some of the column widths in the table to prevent truncated column titles as much as possible
- Adds a `Feature required` detector state - this is to let the user know that a detector is stopped and cannot run until features are added to it

Before: 

Screen Shot 2020-04-27 at 4 56 50 PM

After: 

Screen Shot 2020-04-27 at 4 56 06 PM


public/pages/utils/constants.ts
```diff
@@ -20,6 +20,7 @@ export enum DETECTOR_STATE_COLOR {
   DISABLED = 'subdued',
   INIT = '#0000cc',
```
yizheliu-amazon, line 21: can you try 'primary' for INIT? I recalled 'primary' is blue color as well.

server/routes/ad.ts
```diff
@@ -461,6 +460,16 @@ const getDetectors = async (
...
+      /*
+        If a detector has no features, set to a feature required state
+      */
+      if (
+        detectorState.state === DETECTOR_STATE.DISABLED &&
+        finalDetectors[i].featureAttributes.length === 0
+      ) {
+        detectorState.state = DETECTOR_STATE.FEATURE_REQUIRED;
+      }
     });
```
yizheliu-amazon, line 473: you may encapsulate such logic to a method with appropriate name like `getStateFromRawResponse()` or better name

ohltyler: Sure, good idea. Will add a function in adHelpers.

public/pages/utils/constants.ts
ohltyler: I see, yes. Will change.

PR #55, opened 2020-04-29
add live anomaly reducer

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* add live anomaly reducer

### DIFF
--- a/public/models/interfaces.ts
+++ b/public/models/interfaces.ts
@@ -74,8 +74,8 @@ export type Schedule = {
 
 export type UiFeature = {
   featureType: FEATURE_TYPE;
-  aggregationBy: string;
-  aggregationOf: string;
+  aggregationBy?: string;
+  aggregationOf?: string;
 };
 
 export type UiMetaData = {
@@ -102,6 +102,8 @@ export type Detector = {
   enabled?: boolean;
   enabledTime?: number;
   disabledTime?: number;
+  curState: DETECTOR_STATE;
+  initializationError: string;
 };
 
 export type DetectorListItem = {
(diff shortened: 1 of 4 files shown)


public/models/interfaces.ts
```diff
@@ -102,6 +102,8 @@ export type Detector = {
   enabled?: boolean;
   enabledTime?: number;
   disabledTime?: number;
+  curState: DETECTOR_STATE;
+  initializationError: string;
```
yizheliu-amazon, line 106: ha, I guess those are for `Detector state page`. but not a blocker for approving it.

public/redux/reducers/liveAnomalyResults.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const DETECTOR_LIVE_RESULTS = 'ad/DETECTOR_LIVE_RESULTS';
+
+export interface Anomalies {
+  requesting: boolean;
+  totalLiveAnomalies: number;
+  liveAnomalies: Anomaly[];
+  errorMessage: string;
+}
+export const initialDetectorsState: Anomalies = {
```
yizheliu-amazon, line 34: rename to `initialDetectorLiveResults`?

public/models/interfaces.ts
ylwu-amzn: Yeah, correct

public/redux/reducers/liveAnomalyResults.ts
ylwu-amzn: sure, will change

PR #57, opened 2020-04-29
modify get ad result to return correct anomaly/feature data

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* modify get ad result to return correct anomaly/feature data

### DIFF
--- a/server/models/types.ts
+++ b/server/models/types.ts
@@ -88,26 +88,50 @@ export type GetDetectorsQueryParams = {
   sortField: string;
 };
 
+export type GetAdMonitorsQueryParams = {
+  from: number;
+  size: number;
+  search: string;
+  indices?: string;
+  sortDirection: SORT_DIRECTION;
+  sortField: string;
+};
+
 export type DetectorResultsQueryParams = {
   from: number;
   size: number;
   sortDirection: SORT_DIRECTION;
   sortField: string;
-  range?: object;
+  dateRangeFilter?: DateRangeFilter;
 };
 
 export type AnomalyResult = {
   startTime: number;
   endTime: number;
-  grade: number;
+  plotTime: number;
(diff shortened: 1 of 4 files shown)


server/models/types.ts
```diff
@@ -88,26 +88,50 @@ export type GetDetectorsQueryParams = {
...
 };
 
 export type AnomalyResult = {
   startTime: number;
   endTime: number;
-  grade: number;
+  plotTime: number;
+  anomalyGrade: number;
   confidence: number;
 };
```
ohltyler, line 114: do these changes break code anywhere else that uses these types?

ylwu-amzn: No, this only used by `getAnomalyResults`

server/routes/utils/adHelpers.ts
```diff
@@ -129,25 +129,21 @@ export const anomalyResultMapper = (anomalyResults: any[]): AnomalyResults => {
...
         rest.anomalyGrade != null && rest.anomalyGrade > 0
-          ? Number.parseFloat(rest.confidence).toFixed(3)
+          ? Number.parseFloat(rest.confidence).toFixed(2)
           : 0,
       startTime: rest.dataStartTime,
       endTime: rest.dataEndTime,
-      plotTime:
-        rest.dataStartTime +
-        Math.floor((rest.dataEndTime - rest.dataStartTime) / 2),
+      plotTime: rest.dataEndTime,
```
yizheliu-amazon, line 140: just want to confirm, should we use data end time or data start time as plot time? Dashboard Live chart uses data start time. I guess we may keep consistency.

ylwu-amzn: How about we use data end time? If we use start time, hard to compare alerts history and anomalies on AD result chart.

PR #59, opened 2020-04-29
fix empty ad result; set max monitor size

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* fix empty ad result
* set max monitor size

### DIFF
--- a/server/utils/constants.ts
+++ b/server/utils/constants.ts
@@ -54,3 +54,5 @@ export enum AD_DOC_FIELDS {
   ERROR = 'error',
   INDICES = 'indices',
 }
+
+export const MAX_MONITORS = 1000;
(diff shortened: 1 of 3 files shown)


server/utils/constants.ts
```diff
@@ -54,3 +54,5 @@ export enum AD_DOC_FIELDS {
   ERROR = 'error',
   INDICES = 'indices',
 }
+
+export const MAX_MONITORS = 1000;
```
yizheliu-amazon, line 58: you may put it [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/development/public/utils/constants.ts#L46)

ylwu-amzn: Thanks. Prefer to keep this as server size constraints currently. As it's used only by alerting routes logic from service side.

PR #79, opened 2020-05-01
don't display legend value

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* don't display legend value


ylwu-amzn:
> Can you add 1 more line [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx#L317) as well?

Will add TODO in next PR

Issue #68, opened 2020-04-30
Can't set 0 in detector filter


ylwu-amzn:
![Screen Shot 2020-04-29 at 9 43 27 PM](https://user-images.githubusercontent.com/49084640/80674601-4315fb00-8a67-11ea-81bf-deb675ebcf89.png)

PR #87, opened 2020-05-01
Fix issue that 0 cannot be set in detector filter. Issue: #68

yizheliu-amazon:
*Issue #, if available:*
#68 
*Description of changes:*
Fix issue that 0 cannot be set in detector filter. Issue: #68

Before fix screenshot is in Issue: #68

After fix:

Screen Shot 2020-05-01 at 2 17 02 PM

### DIFF
--- a/public/utils/utils.tsx
+++ b/public/utils/utils.tsx
@@ -42,6 +42,7 @@ export const isInvalid = (name: string, form: any) =>
 export const getError = (name: string, form: any) => get(form.errors, name);
 
 export const required = (val: any): string | undefined => {
+  // if val is number, skip check as number value already exists
   return typeof val !== 'number' && !val ? 'Required' : undefined;
 };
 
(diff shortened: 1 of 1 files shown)


public/utils/utils.tsx
```diff
@@ -42,7 +42,8 @@ export const isInvalid = (name: string, form: any) =>
 export const getError = (name: string, form: any) => get(form.errors, name);
 
 export const required = (val: any): string | undefined => {
-  return !val ? 'Required' : undefined;
+  // if val is number, skip check as number value already exists
+  return typeof val !== 'number' && !val ? 'Required' : undefined;
```
ylwu-amzn, line 46: Can you verify other pages which using this method like edit feature page?

yizheliu-amazon: sure. already verified.

PR #103, opened 2020-05-06
Fix unit test workflow

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR fixes the issue of bootstrapping the plugin in the GitHub runner for our test workflow. By running `yarn kbn bootstrap` a second time, the `node_modules` directory is created and can be used by Yarn to properly bootstrap.

This also removes snapshots that contain converted timestamp information - this causes tests to fail if the tests are run in different timezones (ex: UTC timezone in GitHub runner).

After changes: all UT should run and pass:

Screen Shot 2020-05-06 at 11 23 05 AM

Note: occasionally some tests will time out, lots of tests output errors that need to be addressed later on. Lots of cleaning up and additional tests are needed to increase our code coverage as well.

### DIFF
--- a/public/pages/Dashboard/Components/utils/DashboardHeader.tsx
+++ b/public/pages/Dashboard/Components/utils/DashboardHeader.tsx
@@ -40,7 +40,6 @@ export const DashboardHeader = (props: DashboardHeaderProps) => {
             <EuiButton
               fill
               href={`${PLUGIN_NAME}#${APP_PATH.CREATE_DETECTOR}`}
-              target="_blank"
               data-test-subj="add_detector"
             >
               Create detector
--- a/public/pages/Dashboard/utils/utils.tsx
+++ b/public/pages/Dashboard/utils/utils.tsx
@@ -343,10 +343,7 @@ export const anomalousDetectorsStaticColumn = [
     truncateText: false,
     textOnly: true,
     render: (name: string, detector: Detector) => (
-      <EuiLink
-        href={`${PLUGIN_NAME}#/detectors/${detector.id}/results`}
-        target="_blank"
-      >
+      <EuiLink href={`${PLUGIN_NAME}#/detectors/${detector.id}/results`}>
         {name}
       </EuiLink>
     ),


public/pages/DetectorDetail/containers/__tests__/DetectorDetail.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('<DetectorDetail /> spec', () => {
   describe('detector detail', () => {
     test('renders detector detail component', () => {
       const { container } = renderWithRouter(detectorId);
-      expect(container).toMatchSnapshot();
+      //expect(container).toMatchSnapshot();
```
ylwu-amzn, line 69: Local test can pass this UT. Do you know why it fail in Github UT workflow?

.github/workflows/unit-tests-workflow.yml
```diff
@@ -39,10 +39,17 @@ jobs:
         with:
           path: kibana/plugins/anomaly-detection-kibana-plugin
       - name: Bootstrap the plugin
+        continue-on-error: true
+        run: |
+          cd kibana/plugins/anomaly-detection-kibana-plugin
+          yarn kbn bootstrap
+      - name: Bootstrap the plugin again
```
ylwu-amzn, line 46: Any reason to bootstrap the plugin twice?

public/pages/DetectorDetail/containers/__tests__/DetectorDetail.test.tsx
ohltyler: Github runners run in UTC timezone - when the test runs against the snapshot (which has a converted timestamp time) the test will fail because the converted timestamp in the runner env is different. Ex: tests were failing before here because of a 7-hour offset, which is the difference between PST and UTC

ohltyler: Not that this would cause our tests to fail if anyone runs our tests in any timezone other than PST

ylwu-amzn: Got it. Then we should skip this first. Let's find way to fix these later. Can you create an issue to track this?

.github/workflows/unit-tests-workflow.yml
ohltyler: This is the issue I've been trying to figure out for several days. Could be a bug with Yarn as well. Problem is identical on my local and in github runner. Basically, `yarn kbn bootstrap` is tries to reference the created `node_modules` directory before created. After it runs once, it is created and bootstrap can install all dependencies correctly.

public/pages/DetectorDetail/containers/__tests__/DetectorDetail.test.tsx
ohltyler: What do you mean by skipping? Want me to close PR and create issues for these changes?

ylwu-amzn: Not sure if Jest have some annotation to skip some test cases. That's ok we comment out the test case currently. 
No, we can just create a issue like "fix snapshot test with converted timestamp". No need to close this PR.

ohltyler: Agreed, yeah. Will create issue. Also yeah, let me change to skipping the test. That way we still preserve the snapshots that can be changed later.

PR #104, opened 2020-05-06
Fix test workflow

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR fixes the issue of bootstrapping the plugin in the GitHub runner for our test workflow. By running `yarn kbn bootstrap` a second time, the `node_modules` directory is created and can be used by Yarn to properly bootstrap.

This also skips tests that compare against snapshots that contain converted timestamp information - this causes tests to fail if the tests are run in different timezones (ex: UTC timezone in GitHub runner). Will create an issue to follow up with this.

After changes, all UT should run and pass.

**Note:** occasionally some tests will time out, lots of tests output errors that need to be addressed later on. Lots of cleaning up and additional tests are needed to increase our code coverage as well.

### DIFF
--- a/.github/workflows/unit-tests-workflow.yml
+++ b/.github/workflows/unit-tests-workflow.yml
@@ -39,10 +39,17 @@ jobs:
         with:
           path: kibana/plugins/anomaly-detection-kibana-plugin
       - name: Bootstrap the plugin
+        continue-on-error: true
+        run: |
+          cd kibana/plugins/anomaly-detection-kibana-plugin
+          yarn kbn bootstrap
+      - name: Bootstrap the plugin again
+        if: ${{ always() }}
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-tests-workflow.yml
```diff
@@ -39,10 +39,17 @@ jobs:
         with:
           path: kibana/plugins/anomaly-detection-kibana-plugin
       - name: Bootstrap the plugin
+        continue-on-error: true
+        run: |
+          cd kibana/plugins/anomaly-detection-kibana-plugin
+          yarn kbn bootstrap
```
yizheliu-amazon, line 45: this step looks like to be attempting bootstrap, maybe you can change the name to "Attempt bootstrap"

ohltyler: sure, yeah

PR #113, opened 2020-05-08
Fix live chart pulling

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*

- Improve performance of fetching anomalies when there are too many anomaly results
- Fix detector stopped status message


ohltyler:
Is there any broken UT we can easily fix?

yizheliu-amazon:
> Is there any broken UT we can easily fix?

yeah. working on the fix now. will send out new commit.

PR #116, opened 2020-05-08
fix live chart bar width problem

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* fix live chart bar width problem

### DIFF
--- a/public/pages/utils/anomalyResultUtils.ts
+++ b/public/pages/utils/anomalyResultUtils.ts
@@ -111,7 +111,7 @@ export const prepareDataForChart = (
   data: any[],
   dateRange: DateRange,
   interval: number = 1,
-  getFloorPlotTime?: any,
+  getFloorPlotTime?: any
 ) => {
   if (!data || data.length === 0) {
     return [];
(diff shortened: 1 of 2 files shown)


public/pages/utils/anomalyResultUtils.ts
```diff
@@ -124,32 +124,42 @@ export const prepareDataForChart = (
...
+      confidence: null,
+      anomalyGrade: null,
+    });
+    anomalies.unshift({
+      startTime: dateRange.endDate,
+      endTime: dateRange.endDate,
+      plotTime: dateRange.endDate,
+      confidence: null,
+      anomalyGrade: null,
+    });
```
yizheliu-amazon, line 162: will X Axis automatically shrink if only start/end data points are inserted? I recall I got such issue if plotting data only has start/end data.

yizheliu-amazon: nvm. just realized you insert real data in above code

ylwu-amzn: Good question, let me test locally.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -124,32 +124,42 @@ export const prepareDataForChart = (
...
-    startTime: dateRange.endDate,
-    endTime: dateRange.endDate,
-    plotTime: interval
-      ? dateRange.endDate + MIN_IN_MILLI_SECS * interval
-      : dateRange.endDate,
-    confidence: null,
-    anomalyGrade: null,
-  });
-  if (getFloorPlotTime) {
+  if (getFloorPlotTime) { // we need to get floor plot time for bar chart
```
yizheliu-amazon, line 127: question: In which cases `getFloorPlotTime ` is undefined?

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: I think for your case it should be fine. Anomalies can only be empty when detector is not running, and in that case, we hide live chart.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: As discussed offline, there exists case where `getFloorPlotTime ` is not needed, aka LineSeries. I am good with that.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -124,24 +124,53 @@ export const prepareDataForChart = (
...
+      anomalies.length > 0
+        ? anomalies[anomalies.length - 1].plotTime
+        : getFloorPlotTime(dateRange.startDate);
+    let endTime =
+      anomalies.length > 0
+        ? anomalies[0].plotTime
+        : getFloorPlotTime(dateRange.endDate);
+
+    while (endTime < dateRange.endDate) {
+      endTime += MIN_IN_MILLI_SECS * interval;
```
yizheliu-amazon, line 145: add comment saying why we need this, such as: "make sure the end of plot time is big enough to cover NOW annotation, which is at dateRange.endDate"

ylwu-amzn: sure

PR #122, opened 2020-05-09
fix live chart time range;add custom anomaly tooltip

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* Fix wrong bar width cause by X Axis time range

## Before
X Axis shows 1 hour, which is wrong, we should show 6 hours.
Screen Shot 2020-05-09 at 1 06 29 AM

## After
Screen Shot 2020-05-09 at 4 01 45 AM

* Add custom tooltip for live anomaly

## Before
Screen Shot 2020-05-09 at 3 59 46 AM

## After
Screen Shot 2020-05-09 at 3 56 18 AM

### DIFF
--- a/public/pages/utils/anomalyResultUtils.ts
+++ b/public/pages/utils/anomalyResultUtils.ts
@@ -107,11 +107,43 @@ const sampleMaxAnomalyGrade = (anomalies: any[]): any[] => {
   return sampledAnomalies;
 };
 
+export const prepareDataForLiveChart = (
+  data: any[],
+  dateRange: DateRange,
+  interval: number
+) => {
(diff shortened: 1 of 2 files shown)


public/pages/utils/anomalyResultUtils.ts
```diff
@@ -107,11 +107,43 @@ const sampleMaxAnomalyGrade = (anomalies: any[]): any[] => {
   return sampledAnomalies;
 };
 
+export const prepareDataForLiveChart = (
```
yizheliu-amazon, line 110: maybe add comment saying why we need this.

PR #123, opened 2020-05-10
Use original anomaly grade for dashboard; use fixed height for live chart

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Use original anomaly grade for dashboard; use fixed height for live chart

Before the change, live chart/sunburst chart will ignore anomaly grade < 0.005. Since I do `Number(anomaly_grade).toFixed(2)`, and filter out number === 0, any number less than 0.005 will be 0.00 after `toFixed()`. This causes that dashboard shows 1 anomaly less the number on Detector list.

Another change, use fixed height for dashboard live chart so that when loading icon is showing during live chart refreshing, dashboard live chart will shrink.

### DIFF
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -47,7 +47,10 @@ import {
 } from '@elastic/charts';
 import { EuiText, EuiTitle } from '@elastic/eui';
 import React from 'react';
-import { TIME_NOW_LINE_STYLE } from '../utils/constants';
+import {
+  TIME_NOW_LINE_STYLE,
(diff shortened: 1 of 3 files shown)


kaituo:
> Another change, use fixed height for dashboard live chart so that when loading icon is showing during live chart refreshing, dashboard live chart will shrink.

Do you mean dashboard live chart will *not* shrink if we use fixed height?

yizheliu-amazon:
> > Another change, use fixed height for dashboard live chart so that when loading icon is showing during live chart refreshing, dashboard live chart will shrink.
> 
> Do you mean dashboard live chart will _not_ shrink if we use fixed height?
Yes. Shrinking only happens when refreshing, because loading icon shows up during that time, if there is no fixed height.

Issue #91, opened 2020-05-02
Switch to bucket aggregation to improve anomaly distribution performance


ylwu-amzn:
Currently we query raw anomalies which has `anomaly_grade>0` and aggregate by detector. Generally, the anomalies count should be small. But for long time range or edge case like 1k detectors, it may result in long time loading. We can try bucket aggregation to improve performance.

PR #125, opened 2020-05-10
Fix bugs on dashboard live chart

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Fix bug when over 10000 results are hit for dashboard live chart 1st time fetching anomalies, although 1 result is needed;
Use fixed height for loading spinner, instead of live chart panel

### DIFF
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -101,7 +101,8 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
       '30m',
       dispatch,
       -1,
-      1
+      1,
+      true
     );
 
     setHasLatestAnomalyResult(!isEmpty(latestSingleLiveAnomalyResult));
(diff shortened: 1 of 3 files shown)


public/pages/Dashboard/utils/utils.tsx
```diff
@@ -517,7 +522,8 @@ export const getLatestAnomalyResultsByTimeRange = async (
     );
     anomalyResults = [...anomalyResults, ...anomalies];
     from += anomalySize;
-  } while (numResults === MAX_ANOMALIES);
+    numSingleBatchResults = anomalies.length;
+  } while (numSingleBatchResults === MAX_ANOMALIES);
```
ylwu-amzn, line 526: If too many results, is it possible we will do a lot of while loops and result in timeout?

public/pages/Dashboard/utils/utils.tsx
```diff
@@ -571,7 +579,8 @@ export const getLatestAnomalyResultsForDetectorsByTimeRange = async (
...
```
ylwu-amzn, line 583: Same here. May be Timeout ?

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -101,7 +101,8 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
       '30m',
       dispatch,
       -1,
```
kaituo, line 103: Could you add a comment here explaining why we are fetching -1 instead of 0?

public/pages/Dashboard/utils/utils.tsx
yizheliu-amazon: Good point. Thanks for creating the issue #91 . Will fix it after release.

PR #129, opened 2020-05-11
Fix loading bug on anomalies live chart

ohltyler:
*Issue #, if available:*

*Description of changes:*

Adds error checking to anomalies live chart in the case that .opendistro-anomaly-results index does not exist yet. Without error checking, chart was getting stuck in an infinite loading state. This fixes that.

Before:

Screen Shot 2020-05-11 at 10 52 28 AM

After:

Screen Shot 2020-05-11 at 10 51 59 AM

### DIFF
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -96,14 +96,22 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
   const getLiveAnomalyResults = async () => {
     setIsLoadingAnomalies(true);
     // check if there is any anomaly result in last 30mins
-    const latestSingleLiveAnomalyResult = await getLatestAnomalyResultsByTimeRange(
-      searchES,
-      '30m',
-      dispatch,
-      -1,
-      1,
-      true
-    );
+    // need to initially check if there is an error when accessing anomaly results index
+    // in the case that it doesn't exist upon cluster initialization
+    let latestSingleLiveAnomalyResult = [] as any[];
+    try {
+      latestSingleLiveAnomalyResult = await getLatestAnomalyResultsByTimeRange(
+        searchES,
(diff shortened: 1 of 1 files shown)


public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -96,14 +96,22 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
...
-    const latestSingleLiveAnomalyResult = await getLatestAnomalyResultsByTimeRange(
-      searchES,
-      '30m',
-      dispatch,
-      -1,
-      1,
-      true
-    );
+    // need to initially check if there is an error when accessing anomaly results index
+    // in the case that it doesn't exist upon cluster initialization
```
ylwu-amzn, line 100: Good catch. Can you help check if other places have similar issue? Like detector list page?

ohltyler: checked - detector list page doesn't ping anomaly results idx

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -96,14 +96,22 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
...
+    let latestSingleLiveAnomalyResult = [] as any[];
+    try {
+      latestSingleLiveAnomalyResult = await getLatestAnomalyResultsByTimeRange(
+        searchES,
+        '30m',
+        dispatch,
+        -1,
+        1,
+        true
```
yizheliu-amazon, line 109: what if you change this to `false`?

ohltyler: wouldn't that change the functionality of the call? I think we just want to encapsulate any error here (which can definitely be caused by the index not existing yet) and set is loading to false.

PR #130, opened 2020-05-12
Fix callout when error creating detector

ohltyler:
*Issue #, if available:*

*Description of changes:*

When user has reached the limit of detectors, a confusing exception is thrown and shown to the user when trying to create a new detector. This fixes that, and shows a more informative callout message.

Before:

Screen Shot 2020-05-11 at 6 52 28 PM

After:

Screen Shot 2020-05-11 at 6 51 04 PM

### DIFF
--- a/public/pages/createDetector/containers/CreateDetector.tsx
+++ b/public/pages/createDetector/containers/CreateDetector.tsx
@@ -161,7 +161,7 @@ export function CreateDetector(props: CreateADProps) {
       },
     } = props;
     if (isEmpty(detectorName)) {
-      throw 'Detector name can not be empty';
+      throw 'Detector name cannot be empty';
     } else {
       const error = validateName(detectorName);
       if (error) {
--- a/public/pages/createDetector/containers/CreateDetector.tsx
+++ b/public/pages/createDetector/containers/CreateDetector.tsx
@@ -41,7 +41,7 @@ import {
   searchDetector,
   updateDetector,
 } from '../../../redux/reducers/ad';
-import { BREADCRUMBS } from '../../../utils/constants';
+import { BREADCRUMBS, MAX_DETECTORS } from '../../../utils/constants';
 import { getErrorMessage, validateName } from '../../../utils/utils';
 import { DetectorInfo } from '../components/DetectorInfo';
 import { useFetchDetectorInfo } from '../hooks/useFetchDetectorInfo';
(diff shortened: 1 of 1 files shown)


public/pages/createDetector/containers/CreateDetector.tsx
```diff
@@ -119,9 +119,17 @@ export function CreateDetector(props: CreateADProps) {
         `/detectors/${detectorResp.data.response.id}/configurations/`
       );
     } catch (err) {
-      toastNotifications.addDanger(
-        getErrorMessage(err, 'There was a problem creating detector')
+      const resp = await dispatch(
```
ylwu-amzn, line 122: How about we use _count API?

ohltyler: I'm not sure how that would apply here; I'm just leveraging our existing `searchDetector` function. Similar statement used [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/c5bb7c45b707a8d944e2fb9b252c95f77e468c38/public/pages/createDetector/containers/CreateDetector.tsx#L172) and I wanted to keep it consistent.

ohltyler: Wait I see; query .opendistro-anomaly-detectors directly. Let me try

ylwu-amzn: Maybe another easy way is passing detector count to detector creation page. As dashboard and detector list page has already got total detector count.

ylwu-amzn: Oh, sorry, if pass detector count, we can't avoid the case that user open creation page directly by typing url. So we still need to count the total detector on creation page.

ohltyler: I'm concerned it's unnecessary to add a separate function to get the count using the _count API; this would add a lot of code for a function that may not need to be used outside of this case.

ylwu-amzn: Got it. Let's use this now as the performance is acceptable considering we have 1k detectors at most. Let's tune it later.

ohltyler: As discussed offline, will leave as-is for now. Current fix doesn't affect performance regarding showing the callout. Future improvements to this and other parts of code may use _count API.

PR #147, opened 2020-05-15
Fix and clean unit tests

ohltyler:
*Issue #, if available:* #148 

*Description of changes:*

This PR cleans up some of the unit tests:
- fixes a test that can timeout occasionally on macos-latest, causing workflow to fail
- suppresses error output while tests are running so output is more readable
- fixes a few React warnings when rendering certain pages (making sure all items in a list have a key, passing valid props to React components)

On both ubuntu-latest and macos-latest GitHub runners: 

Screen Shot 2020-05-15 at 3 57 21 PM

### DIFF
--- a/public/pages/createDetector/components/DataFilters/SimpleFilter.tsx
+++ b/public/pages/createDetector/components/DataFilters/SimpleFilter.tsx
@@ -100,7 +100,6 @@ export const SimpleFilter = (props: DataFilterProps) => {
               return (
                 <EuiPanel key={index} className="filter-container">
                   <EuiAccordion
-                    key={index}
                     id={'name'}
                     initialIsOpen={true}
                     buttonContent={displayText(filter)}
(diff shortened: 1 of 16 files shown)


public/pages/createDetector/components/DataFilters/SimpleFilter.tsx
```diff
@@ -98,7 +98,7 @@ export const SimpleFilter = (props: DataFilterProps) => {
           <EuiPanel style={darkMode ? {} : lightModeStyles}>
             {values.filters.map((filter: UIFilter, index: number) => {
               return (
-                <EuiPanel className="filter-container">
+                <EuiPanel key={index} className="filter-container">
```
kaituo, line 101: Would the key of the euipanel conflict with euiaccordion since they are both using "index"?

ohltyler: Good catch. React doesn't seem to throw any warnings about this, although it would probably be best to remove the 'key' prop from the `EuiAccordion`, since we only need the key to be at the top level of the component in the list. Thoughts?

PR #152, opened 2020-05-18
tune code style for ad result utils

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* Tune code style for ad result utils

### DIFF
--- a/public/pages/AnomalyCharts/utils/constants.ts
+++ b/public/pages/AnomalyCharts/utils/constants.ts
@@ -88,3 +88,12 @@ export enum LIVE_CHART_CONFIG {
   REFRESH_INTERVAL_IN_SECONDS = 30 * 1000, //poll anomaly result every 30 seconds
   MONITORING_INTERVALS = 60,
 }
+
+export const DEFAULT_ANOMALY_SUMMARY = {
+  anomalyOccurrence: 0,
+  minAnomalyGrade: 0,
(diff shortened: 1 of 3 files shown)


public/pages/AnomalyCharts/utils/constants.ts
```diff
@@ -88,3 +88,12 @@ export enum LIVE_CHART_CONFIG {
   REFRESH_INTERVAL_IN_SECONDS = 30 * 1000, //poll anomaly result every 30 seconds
   MONITORING_INTERVALS = 60,
 }
+
+export const DEFAULT_ANOMALY_SUMMARY = {
+  anomalyOccurrence: 0,
+  minAnomalyGrade: 0,
+  maxAnomalyGrade: 0,
+  minConfidence: 0,
+  maxConfidence: 0,
```
ohltyler, line 97: Small: for these 4 fields (min/max grade/confidence), can we change these to be strings as well, and display something like '-' here instead? Since these fields are irrelevant in this case, it seems more appropriate than just showing 0s.

ylwu-amzn: This is reviewed by UX designer. Will keep this as is currently.

ohltyler: Got it, sounds good.

Issue #151, opened 2020-05-18
Fix loading state of detector detail page


ohltyler:
When loading the detector details page, the intermediate loading state always shows an empty detector name and is in the 'Feature required' state.

While loading:

<img width="285" alt="Screen Shot 2020-05-18 at 1 57 16 PM" src="https://user-images.githubusercontent.com/62119629/82259156-f9a32800-990f-11ea-8307-4b8d645f6b9a.png">

After loading:

<img width="274" alt="Screen Shot 2020-05-18 at 1 57 34 PM" src="https://user-images.githubusercontent.com/62119629/82259189-03c52680-9910-11ea-8b52-d6a249c4e8b1.png">

This should be changed to display something less confusing, and indicate to the user that the detector details are being loaded.

PR #112, opened 2020-05-08
fix blank detector detail page

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*

* fix blank detector detail page

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #155, opened 2020-05-19
Fix detector detail loading state

ohltyler:
*Issue #, if available:* #151 

*Description of changes:*

When loading the detector details page, the intermediate state shows a blank detector name and defaults to the 'Feature required' detector state.

This fixes that by keeping track of when data is being requested and loaded, so an intermediate loading state can be shown. Some logic and code cleanup was also made regarding rendering the detector state at the top of the page.

Before:

Screen Shot 2020-05-18 at 5 00 16 PM

After:

Screen Shot 2020-05-18 at 7 57 58 PM

This PR also makes a minor change to the detector list loading state, where the requesting of info from the backend and the sorting/filtering in the frontend both display a consistent loading page to the user.

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -64,12 +64,12 @@ interface DetectorDetailProps
 const tabs = [
   {
     id: DETECTOR_DETAIL_TABS.RESULTS,
-    name: 'Anomaly Results',
+    name: 'Anomaly results',
     route: DETECTOR_DETAIL_TABS.RESULTS,
   },
   {
     id: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
-    name: 'Detector Configuration',
+    name: 'Detector configuration',
     route: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
   },
 ];
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -52,6 +52,7 @@ import { MonitorCallout } from '../components/MonitorCallout/MonitorCallout';
 import { DETECTOR_DETAIL_TABS } from '../utils/constants';
 import { DetectorConfig } from '../../DetectorConfig/containers/DetectorConfig';
 import { AnomalyResults } from '../../DetectorResults/containers/AnomalyResults';
+import { DETECTOR_STATE_COLOR } from '../../utils/constants';
 
 export interface DetectorRouterProps {
   detectorId?: string;
@@ -67,7 +68,7 @@ const tabs = [
   },
   {
     id: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
-    name: 'Detector configuration',
+    name: 'Detector Configuration',
     route: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
   },
 ];
@@ -89,13 +90,15 @@ interface DetectorDetailModel {
 export const DetectorDetail = (props: DetectorDetailProps) => {
   const dispatch = useDispatch();
   const detectorId = get(props, 'match.params.detectorId', '') as string;
-  const { detector, hasError } = useFetchDetectorInfo(detectorId);
-  const { monitor, fetchMonitorError } = useFetchMonitorInfo(detectorId);
+  const { detector, hasError, isLoadingDetector } = useFetchDetectorInfo(detectorId);
+  const { monitor, fetchMonitorError, isLoadingMonitor } = useFetchMonitorInfo(detectorId);
+
+  const isLoading = isLoadingDetector || isLoadingMonitor;
 
   //TODO: test dark mode once detector configuration and AD result page merged
   const isDark = darkModeEnabled();
 
-  const [detecorDetailModel, setDetectorDetailModel] = useState<
+  const [detectorDetailModel, setDetectorDetailModel] = useState<
     DetectorDetailModel
   >({
     selectedTab: getSelectedTabId(
@@ -128,7 +131,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
 
   const handleSwitchToConfigurationTab = useCallback(() => {
     setDetectorDetailModel({
-      ...detecorDetailModel,
+      ...detectorDetailModel,
       selectedTab: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
     });
     props.history.push(`/detectors/${detectorId}/configurations`);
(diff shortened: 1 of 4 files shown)


public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -271,30 +274,27 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
...
+                ) : detector.curState === DETECTOR_STATE.DISABLED ? (
+                  <EuiHealth color={DETECTOR_STATE_COLOR.DISABLED}>
+                    {detector.disabledTime
+                    ? `Stopped at ${moment(detector.disabledTime).format('MM/DD/YY h:mm A')}`
+                    : 'Detector is stopped'}
                   </EuiHealth>
-                )}
+                ) : detector.curState === DETECTOR_STATE.FEATURE_REQUIRED ? (
+                  <EuiHealth color={DETECTOR_STATE_COLOR.FEATURE_REQUIRED}>Feature required to start the detector</EuiHealth>
+                ) : isLoading ? 'Loading...' : ''}
```
yizheliu-amazon, line 297: one thought: what about using some animation of [EUI](https://elastic.github.io/eui/#/display/loading), such as **Text content**, instead of 'Loading...'? Maybe you can confirm with UX on this.

ohltyler: I agree, that could be cleaner. I will change.

public/pages/createDetector/hooks/useFetchDetectorInfo.ts
```diff
@@ -55,5 +60,6 @@ export const useFetchDetectorInfo = (
   return {
     detector: detector || {},
     hasError: !isEmpty(hasError) && isEmpty(detector),
+    isLoadingDetector: isDetectorRequesting || isIndicesRequesting,
```
yizheliu-amazon, line 63: Not blocker: previously we have seen one [issue](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/112/files) where `isDetectorRequesting` is true while no other request is going, but we haven't figured out what is the root cause. I think for now we may need to keep eye on the DetectorDetail page to see if there is any weird thing after change is deployed to test domain.

ohltyler: I see, ok. When making these changes I've noticed issues with error handling as well. I'm pretty sure it is related to the unhandled Promise rejection thrown [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/routes/ad.ts#L468) when there is an error retrieving a detector from the backend (invalid id, timeout, etc.). I'm thinking that this could be causing the request to never finish, and leave the ad state's `requesting` field to true. 

I will need to write up an issue to document this as well

ohltyler: Example: when changing the url for detector configuration page to an invalid detector id: page is permanently in loading state and chrome devtools shows this: 
Screen Shot 2020-05-18 at 6 07 25 PM

ylwu-amzn: Cool, let's create separate issue to track.

ohltyler: Detailed explanation here: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/157

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -67,7 +69,7 @@ const tabs = [
   },
   {
     id: DETECTOR_DETAIL_TABS.CONFIGURATIONS,
-    name: 'Detector configuration',
+    name: 'Detector Configuration',
```
ylwu-amzn, line 72: `Detector configuration` is correct, we should change `Anomaly Results` to `Anomaly results`

ohltyler: Oh I see, thanks. Will change.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -89,13 +91,15 @@ interface DetectorDetailModel {
 export const DetectorDetail = (props: DetectorDetailProps) => {
   const dispatch = useDispatch();
   const detectorId = get(props, 'match.params.detectorId', '') as string;
-  const { detector, hasError } = useFetchDetectorInfo(detectorId);
-  const { monitor, fetchMonitorError } = useFetchMonitorInfo(detectorId);
+  const { detector, hasError, isLoadingDetector } = useFetchDetectorInfo(detectorId);
+  const { monitor, fetchMonitorError, isLoadingMonitor } = useFetchMonitorInfo(detectorId);
+
+  const isLoading = isLoadingDetector || isLoadingMonitor;
```
ylwu-amzn, line 97: How about we don't show anything is detector is not found or is loading? Show these two tabs may confuse user that they should switch tab to see more.

ylwu-amzn: I didn't know you are working on this issue, so I sent out a PR too. You can refer to this PR: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/156

ylwu-amzn: Should consider `hasError` to handle exceptions like `not found` , `timeout` etc. If any problem occur, we already have code to pop up toast:
```
 useEffect(() => {
    if (hasError) {
      toastNotifications.addDanger('Unable to find detector');
      props.history.push('/detectors');
    }
  }, [hasError]);
```

ohltyler: > How about we don't show anything if detector not found or still loading? Show these two tabs may confuse user that they should switch tab to see more.

Ok, I agree. Will change.

ohltyler: > Should consider `hasError` to handle exceptions like `not found` , `timeout` etc. We already have code to pop up toast and redirect to detector list page.
> 
> ```
>  useEffect(() => {
>     if (hasError) {
>       toastNotifications.addDanger('Unable to find detector');
>       props.history.push('/detectors');
>     }
>   }, [hasError]);
> ```

I agree, I've just tested locally where when I change the URL to an invalid detector id I get in the infinite loading state, which I believe has to do with [this issue](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/157)

public/pages/createDetector/hooks/useFetchDetectorInfo.ts
yizheliu-amazon: cool. thanks for doing that.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ylwu-amzn: Yes, we have some issue for error handling. Let's fix the infinite loading in another PR.

Issue #133, opened 2020-05-13
Fix long initialization issues in anomaly detection


ylwu-amzn:
Detector initialization process needs at least 6 data points for continuous 8 intervals to complete shingle process. If no data or has not enough data, user may experience long initialization period. We should tune the error message to show something like "no data found" or "no enough data". So user can know why initialization takes a long time.

We can query feature data when create detector to make sure there is enough data. And we can add max empty query limit to stop detector.

ylwu-amzn:
We should add real auto refresh. Currently the initialization state just show a loading spinner, will not send request to query latest state.

yizheliu-amazon:
Currently we just show message in UI if long initialization is found, but don't stop detector. In the long term, I think we should stop detector from backend if we identify long initialization, and treat it as initialization failure if possible.

epotocko:
Could existing data in the ElasticSearch index be used to initialize the detector? From my usage, I've only seen records created while the detector is running used for anomaly detection. I would have expected it to use all data available in the index or at least a subset of recent data.

ylwu-amzn:
> Could existing data in the ElasticSearch index be used to initialize the detector? From my usage, I've only seen records created while the detector is running used for anomaly detection. I would have expected it to use all data available in the index or at least a subset of recent data.

Good question. @wnbts , can you help explain?

wnbts:
Hi @epotocko , the current "initialization" page is misleading/overloaded. The system does use existing data to complete model training on the backend. However, the real time data stream might be a root cause. If you use the rest api, do you see produced results?

epotocko:
Which API call are you referring to? Shortly after creating and starting a detector with 4+ months of data points and a 60 minute detector interval:
The _preview API returns: { "anomaly_result" : [ ]..........
The _profile API returns: { "state" : "INIT" }

I can reproduce this consistently with different data sets. The _profile API will always return an INIT state until enough "new" records are received.

wnbts:
@epotocko  I was referring to get anomaly results api. Do you see anomaly results and feature values since the start of the detector? And show maybe 10~20 examples of recent results if you can to see if there are any errors. That will help us see whether it's a data stream issue or a system issue. Thanks!

epotocko:
@wnbts Immediately after creating the detector, the get anomaly results api returns 0 hits. I checked about an hour later and there was one hit with the error "No full shingle in current detection window". 

[...]

PR #159, opened 2020-05-19
Add proper message in case of long initialization. Issue:#133

yizheliu-amazon:
*Issue #, if available:*
#133 
*Description of changes:*
Add proper message in case of long initialization. Initialization is considered as overtime if current initialization > SHINGLE_SIZE * detectionInterval + enabledTime.

No full shingle error:
Screen Shot 2020-05-18 at 9 03 32 PM
Screen Shot 2020-05-18 at 9 04 39 PM

No data in current time window error:
Screen Shot 2020-05-18 at 9 33 36 PM
Screen Shot 2020-05-18 at 9 33 58 PM

### DIFF
--- a/public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
+++ b/public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
@@ -52,8 +52,8 @@ export const DetectorInitializing = (props: DetectorInitializingProps) => {
             ? [
                 <p>
                   Based on your latest update to the detector configuration, the
-                  detector is collecting sufficient data to generate accurate
-                  real-time anomalies.
+                  detector is collecting data to generate accurate real-time
+                  anomalies.
                 </p>,
                 <p>
                   The longer the detector interval is, the more time this will
--- a/public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
+++ b/public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
@@ -16,6 +16,14 @@ import React from 'react';
 import { EuiButton, EuiEmptyPrompt, EuiLoadingSpinner } from '@elastic/eui';
 import { Fragment } from 'react';
 import { Detector } from '../../../../models/interfaces';
+import {
+  getDetectorInitializationInfo,
+  IS_INIT_OVERTIME_FIELD,
+  INIT_DETAILS_FIELD,
+  INIT_ERROR_MESSAGE_FIELD,
+  INIT_ACTION_ITEM_FIELD,
+} from '../../utils/utils';
+import { get } from 'lodash';
 
 export interface DetectorInitializingProps {
   detector: Detector;
@@ -23,6 +31,12 @@ export interface DetectorInitializingProps {
 }
 
 export const DetectorInitializing = (props: DetectorInitializingProps) => {
+  const initializationInfo = getDetectorInitializationInfo(props.detector);
+  const isInitOvertime = get(initializationInfo, IS_INIT_OVERTIME_FIELD, false);
+  const initDetails = get(initializationInfo, INIT_DETAILS_FIELD, {});
+  const initErrorMessage = get(initDetails, INIT_ERROR_MESSAGE_FIELD, '');
+  const initActionItem = get(initDetails, INIT_ACTION_ITEM_FIELD, '');
+
   return (
     <EuiEmptyPrompt
       style={{ maxWidth: '75%' }}
@@ -34,14 +48,24 @@ export const DetectorInitializing = (props: DetectorInitializingProps) => {
       }
       body={
         <Fragment>
-          <p>
-            Based on your latest update to the detector configuration, the
-            detector is collecting sufficient data to generate accurate
-            real-time anomalies.
-          </p>
-          <p>
-            The longer the detector interval is, the more time this will take.
-          </p>
+          {!isInitOvertime
+            ? [
+                <p>
+                  Based on your latest update to the detector configuration, the
+                  detector is collecting sufficient data to generate accurate
+                  real-time anomalies.
+                </p>,
+                <p>
+                  The longer the detector interval is, the more time this will
+                  take.
+                </p>,
+              ]
+            : [
+                <p>
+                  {`Detector initialization is not complete because ${initErrorMessage}.`}
+                </p>,
+                <p>{`${initActionItem}`}</p>,
+              ]}
         </Fragment>
       }
       actions={
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -33,6 +33,13 @@ import { BREADCRUMBS, DETECTOR_STATE } from '../../../utils/constants';
 import { AnomalyResultsLiveChart } from './AnomalyResultsLiveChart';
 import { AnomalyHistory } from './AnomalyHistory';
 import { DetectorStateDetails } from './DetectorStateDetails';
+import {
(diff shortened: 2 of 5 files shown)


public/pages/DetectorResults/utils/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+    [INIT_ERROR_MESSAGE_FIELD]: '',
+    [INIT_ACTION_ITEM_FIELD]: '',
+  };
+  if (!detector.initializationError) {
+    return result;
+  }
+  if (detector.initializationError.includes(NO_FULL_SHINGLE_ERROR_MESSAGE)) {
+    result[INIT_ERROR_MESSAGE_FIELD] = 'collected data is insufficient';
+    result[INIT_ACTION_ITEM_FIELD] =
+      DETECTOR_INIT_FAILURES.NO_TRAINING_DATA.actionItem;
```
ohltyler, line 72: small: should there be a separate `NO_FULL_SHINGLE` enum instead of `NO_TRAINING_DATA` here? It feels like `NO_TRAINING_DATA` should be reserved for the `NO_DATA_IN_WINDOW_ERROR_MESSAGE` below this

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
```diff
@@ -34,14 +48,24 @@ export const DetectorInitializing = (props: DetectorInitializingProps) => {
...
-          </p>
-          <p>
-            The longer the detector interval is, the more time this will take.
-          </p>
+          {!isInitOvertime
+            ? [
+                <p>
+                  Based on your latest update to the detector configuration, the
+                  detector is collecting sufficient data to generate accurate
+                  real-time anomalies.
```
kaituo, line 56: This does not seem true.  The detector may not receive any data in its shingle before isInitOverTime become true..

public/pages/DetectorResults/utils/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+  return result;
+};
+
+const isDetectorInitOverTime = (currentTime: Moment, detector: Detector) => {
+  return (
+    detector &&
+    detector.curState === DETECTOR_STATE.INIT &&
+    detector.initializationError &&
+    !detector.initializationError.includes(NO_RCF_MODEL_ERROR_MESSAGE) &&
```
ylwu-amzn, line 49: If no full shingle data, will the error message be `NO_RCF_MODEL_ERROR_MESSAGE `?

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
```diff
@@ -34,14 +48,24 @@ export const DetectorInitializing = (props: DetectorInitializingProps) => {
...
+                </p>,
+                <p>
+                  The longer the detector interval is, the more time this will
+                  take.
+                </p>,
+              ]
+            : [
+                <p>
+                  {`Detector initialization is not complete because ${initErrorMessage}.`}
```
kaituo, line 65: There can be grammar error. Some of initErrorMessage has verb and some don't.  How about replace because with a colon?

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -93,7 +106,9 @@ export function AnomalyResults(props: AnomalyResultsProps) {
                       title={
                         isDetectorUpdated
                           ? 'There are change(s) to the detector configuration after the detector is stopped.'
-                          : 'The detector is being re-initialized based on the latest configuration changes.'
+                          : !isInitOvertime
+                          ? 'The detector is being re-initialized based on the latest configuration changes.'
+                          : `Detector initialization is not complete because ${initErrorMessage}.`
```
kaituo, line 111: because may cause grammar error.

public/pages/DetectorResults/utils/utils.ts
yizheliu-amazon: I think for now we don't treat long initialization as a initialization failure case, instead we think it is somewhere between initialization and initialization failure, same as `No_FULL_SHINGLE` . That is why I did not add `No_FULL_SHINGLE`  to `DETECTOR_INIT_FAILURES`. 

And regarding the `actionItem`, I use NO_TRAINING_DATA's actionItem as I think it can work for `No_FULL_SHINGLE` case, because both `NO_TRAINING_DATA` and `No_FULL_SHINGLE` has a common point: no enough data is collected, while `NO_DATA_IN_WINDOW_ERROR_MESSAGE`'s point is no data at all. Besides, `NO_TRAINING_DATA`'s actionItem is to let user ingest enough data or adjust detectionInterval, this can help solve `No_FULL_SHINGLE` issue as well.

public/pages/DetectorResults/utils/utils.ts
yizheliu-amazon: In the normal case, usually after 6 interval, NO_RCF_MODEL_ERROR_MESSAGE will show up, and model training is started. This check here is to avoid breaking normal case.

public/pages/DetectorResults/utils/utils.ts
ohltyler: I see, thanks for the explanation. That makes sense.

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
yizheliu-amazon: This message is shown when detector is just created/started and being initialized. 

I think this message's point is: when initialization is started, and before `isInitOverTime` becomes true, _no matter data is collected or not_, we show it to user to let user know detector is collecting data.

If no data is collected, and it is overtime, then we show message below to let user know the reason and action item; if it is completed, we just show AnomalyResult page instead.

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
yizheliu-amazon: The only 2 initErrorMessages have verb:
`collected data is insufficient`: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/159/files#diff-bf4a77e66d4a8caf187ddf3925f5d5c2R70
`no data exists in current time window`: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/159/files#diff-bf4a77e66d4a8caf187ddf3925f5d5c2R76

Also, I am waiting for response from Ashwin on the wording. I will change it later after he reviews it.

public/pages/DetectorResults/containers/AnomalyResults.tsx
yizheliu-amazon: Same comment above.

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
kaituo: I see.  I looked at public/pages/DetectorResults/utils/constants.ts.  You can leave it as it is.

public/pages/DetectorResults/components/DetectorState/DetectorInitializing.tsx
kaituo: Understood.  I feel "sufficient" is not correct as we are not certain.

yizheliu-amazon: Yeah. You are right. I will remove `sufficient` to avoid confusion.

yizheliu-amazon:
Have asked Ashwin to take a look. no comments from him.

Issue #162, opened 2020-05-19
Blank anomaly result page when initialization failed


ylwu-amzn:
Re-initialization failed, AD result page shows nothing.

![Screen Shot 2020-05-19 at 11 53 42 AM](https://user-images.githubusercontent.com/49084640/82366764-94127280-99c7-11ea-8e27-f21454f269d8.png)

Check AD result:
```
{
    "detector_id" : "yNpUXXEBPdSnBWCFN-zN",
    "data_start_time" : 1589910474476,
    "data_end_time" : 1589910534476,
    "feature_data" : [ ],
    "execution_start_time" : 1589910534476,
    "execution_end_time" : 1589910534505,
    "error" : "Stopped detector as job failed consecutively for more than 3 times: We might have bugs."
  }

{
    "detector_id" : "yNpUXXEBPdSnBWCFN-zN",
    "data_start_time" : 1589910414476,
    "data_end_time" : 1589910474476,
    "feature_data" : [ ],
    "execution_start_time" : 1589910474476,
    "execution_end_time" : 1589910474505,
    "error" : "We might have bugs."
  }

{
    "detector_id" : "yNpUXXEBPdSnBWCFN-zN",
    "data_start_time" : 1589910354476,
    "data_end_time" : 1589910414476,
    "feature_data" : [ ],
    "execution_start_time" : 1589910414476,
    "execution_end_time" : 1589910414536,
    "error" : "We might have bugs."
  }

{
    "detector_id" : "yNpUXXEBPdSnBWCFN-zN",
    "data_start_time" : 1589910294476,
    "data_end_time" : 1589910354476,
    "feature_data" : [ ],
    "execution_start_time" : 1589910354476,
    "execution_end_time" : 1589910354514,
    "error" : "We might have bugs."
  }
```

yizheliu-amazon:
"We might have bugs" is corresponding to `Unexpected failure`, we can change this [line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/DetectorResults/containers/DetectorStateDetails.tsx#L58) to
```
    case DETECTOR_STATE.INIT_FAILURE:
    case DETECTOR_STATE.UNEXPECTED_FAILURE:
      const failureDetail = getInitFailureMessageAndActionItem(
        detector.initializationError
      );
...
```

yizheliu-amazon:
Backend error stack trace:
```
[...]

PR #165, opened 2020-05-19
Handle detector init unexpected failure + detector unknown state. Issues: #162

yizheliu-amazon:
*Issue #, if available:*
#162 
*Description of changes:*
Handle detector init unexpected failure case + detector unknown state gracefully.

Detector init unexpected failure:
Screen Shot 2020-05-19 at 3 35 51 PM

detector unknown state:(ideally, we should not reach this state)
Screen Shot 2020-05-19 at 3 59 46 PM

### DIFF
--- a/public/pages/DetectorResults/components/DetectorState/DetectorUnknownState.tsx
+++ b/public/pages/DetectorResults/components/DetectorState/DetectorUnknownState.tsx
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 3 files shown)


public/pages/DetectorResults/components/DetectorState/DetectorUnknownState.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const DetectorUnknownState = (props: DetectorUnknownStateProps) => {
+  return (
+    <EuiEmptyPrompt
+      style={{ maxWidth: '75%' }}
+      title={
+        <div>
+          <EuiIcon type="alert" size="l" color="danger" />
+
+          <h2>The detector is in unknown state</h2>
```
ylwu-amzn, line 33: Should we add a new unknown state? Or just use initialization failutre?

yizheliu-amazon: I add this just for worst case scenario. Theoretically user should **never** reach this state.

It can only be shown to user if there is bug in our code. It could be new edge case we fail to cover, or some logic bug which leads to unexpected state. Such issues could be related to initialization, or config update, or any other bug on AD Kibana side. That's why I don't use initialization failure.

Issue #157, opened 2020-05-19
Fix error handling when retrieving detectors from backend


ohltyler:
Currently, if rendering a page with an invalid detector id, issues come up on a few different pages. For example, if an invalid id is set in the URL for the detector results page, the page is stuck in an infinite loading state. If there is a high number of detectors and retrieving them causes a request to time out, the dashboard and detector list pages display as if there are no existing detectors.

Possible causes: this [line of code here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/routes/ad.ts#L468) which can throw an unhandled Promise rejection if there is an issue retrieving a detector from the backend (invalid id, request timeout, etc.). This error can be seen in chrome devtools when rendering the page: 
![Screen Shot 2020-05-18 at 6 07 25 PM](https://user-images.githubusercontent.com/62119629/82274232-6c25ff00-9934-11ea-91ce-6101f40c9298.png)

Possible reason for infinite loading: state set to requesting -> request made -> uncaught exception thrown -> state is stuck

Possible reason for saying there are 0 detectors: default state of 0 detectors -> request made -> uncaught exception thrown -> state is stuck

Investigate this more and, if fixed, test out and/or fix the error handling on some of the pages.

ohltyler:
Update:

Infinite loading state: caused by undefined issue when getting the error message when `getDetector()` fails (line [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/reducers/ad.ts#L97))

0 detectors on dashboard and detector issue: caused by error in `getDetectorList()`, which causes the current list of detectors to stay at 0. Error not propagated/shown on these pages.

PR #170, opened 2020-05-21
Fix error handling when getting detector and detector list

ohltyler:
*Issue #, if available:* #157 

*Description of changes:*

Fixes error handling when there are errors (1) getting a detector and (2) getting detector list:
- Fixes error message in `getDetector()` to prevent exception which leads to infinite loading state on detector details pages
- Fixes error handling in `getDetectorList()` which will catch exceptions if there's errors in retrieving detector state for any detectors
- Propagate errors getting detector list to dashboard and detector list page via toast notification (will confirm with UX how to handle this)
- Refactors error messages in the ad reducer to be unique for each function. This is necessary because during detector deletion, a `getDetector()` error is thrown. We don't want to propagate this message to user on detector list page, which is where UI is redirected after deletion. Detector list / dashboard pages should only be concerned about errors retrieving all detectors, not a leftover error message caused by a different page.

**Infinite loading state (before)**
Screen Shot 2020-05-21 at 12 24 06 PM

**Infinite loading state (after)**
Screen Shot 2020-05-21 at 12 24 49 PM

**Error getting detector list (before)**
Screen Shot 2020-05-21 at 12 29 55 PM
Screen Shot 2020-05-21 at 12 30 09 PM

**Error getting detector list (after)**
Screen Shot 2020-05-20 at 4 45 15 PM
Screen Shot 2020-05-20 at 4 44 55 PM

### DIFF
--- a/public/pages/Dashboard/Container/DashboardOverview.tsx
+++ b/public/pages/Dashboard/Container/DashboardOverview.tsx
@@ -32,6 +32,8 @@ import {
 } from '@elastic/eui';
 //@ts-ignore
 import chrome from 'ui/chrome';
+// @ts-ignore
+import { toastNotifications } from 'ui/notify';
 import { AnomalousDetectorsList } from '../Components/AnomalousDetectorsList';
 import {
   GET_ALL_DETECTORS_QUERY_PARAMS,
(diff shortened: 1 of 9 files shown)


ohltyler:
Update regarding `getDetectorList()` errors: rather than displaying the pages and show 0 detectors, UX has decided to display a different page telling the user about the error and steps to fix (refresh page, etc.). There will be a mockup soon.

ohltyler:
Will close and split PR into 2

Issue #172, opened 2020-05-21
Handle infinite loading state on detector detail pages


ohltyler:
Currently, if a bad detector id is passed to the detector details pages (detector results and detector configuration pages), an unhandled exception is thrown and the page gets stuck in an infinite loading state:

<img width="721" alt="Screen Shot 2020-05-21 at 12 24 06 PM" src="https://user-images.githubusercontent.com/62119629/82610848-045df700-9b74-11ea-91d6-59b6d644d046.png">

Handle gracefully and redirect to detector list page.

PR #173, opened 2020-05-21
Handle error on detector detail pages

ohltyler:
*Issue #, if available:* #172 

*Description of changes:*

Handles errors when an invalid detector id is passed to the detector detail pages (results and configuration pages). This could be caused by changing the url directly, or refreshing a stale page if the detector has been deleted in another window.

Before (infinite loading state):
Screen Shot 2020-05-21 at 12 24 06 PM

After (redirects to detector list page, toast displaying error):
Screen Shot 2020-05-21 at 12 24 49 PM

### DIFF
--- a/public/redux/reducers/ad.ts
+++ b/public/redux/reducers/ad.ts
@@ -94,7 +94,7 @@ const reducer = handleActions<Detectors>(
       FAILURE: (state: Detectors, action: APIErrorAction): Detectors => ({
         ...state,
         requesting: false,
-        errorMessage: action.error.data.error,
+        errorMessage: action.error,
       }),
     },
     [START_DETECTOR]: {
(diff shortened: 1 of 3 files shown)


public/redux/reducers/ad.ts
```diff
@@ -94,7 +94,7 @@ const reducer = handleActions<Detectors>(
       FAILURE: (state: Detectors, action: APIErrorAction): Detectors => ({
         ...state,
         requesting: false,
-        errorMessage: action.error.data.error,
+        errorMessage: action.error,
```
ohltyler, line 97: This line here was causing an exception. `data` was undefined, which caused the error to be thrown. Because of this, the ad state never got out of `requesting` and was preventing the detector details pages from handling the error and redirecting to the detector list page.

Issue #176, opened 2020-05-22
Infinite loading on detector list page on cluster initialization


ohltyler:
When a brand new cluster is started, (.opendistro-anomaly-detectors doesn't exist yet) there is an uncaught exception on the detector list page when trying to retrieve all detectors from the backend. This is only triggered if a user changes any filtering/sorting/searching etc. on the page. This should be caught, and the loading state should be set to false.

Screenshot:
<img width="1392" alt="Screen Shot 2020-05-22 at 10 12 35 AM" src="https://user-images.githubusercontent.com/62119629/82692823-3e330a00-9c15-11ea-82fd-3f0a3f94f285.png">

PR #177, opened 2020-05-22
Fix detector list infinite loading on cluster initialization

ohltyler:
*Issue #, if available:* #176 

*Description of changes:*

On cluster initialization (when .opendistro-anomaly-detectors index doesn't exist yet) there is an uncaught exception on the detector list page if a user tries changing any filters / searching / sorting etc., leaving the page in an infinite loading state.
This fixes that by fixing the error handling when making the backend call.

Before:
Screen Shot 2020-05-22 at 10 12 35 AM

After:
Screen Shot 2020-05-22 at 10 11 58 AM

### DIFF
--- a/public/pages/DetectorsList/List/List.tsx
+++ b/public/pages/DetectorsList/List/List.tsx
@@ -89,11 +89,19 @@ export const DetectorList = (props: ListProps) => {
     (state: AppState) => state.elasticsearch
   );
 
-  const isRequestingFromES = useSelector((state: AppState) => state.ad.requesting);
+  const isRequestingFromES = useSelector(
+    (state: AppState) => state.ad.requesting
+  );
 
-  const [selectedDetectors, setSelectedDetectors] = useState([] as DetectorListItem[]);
-  const [detectorsToDisplay, setDetectorsToDisplay] = useState([] as DetectorListItem[]);
-  const [isLoadingFinalDetectors, setIsLoadingFinalDetectors] = useState<boolean>(true);
+  const [selectedDetectors, setSelectedDetectors] = useState(
+    [] as DetectorListItem[]
+  );
+  const [detectorsToDisplay, setDetectorsToDisplay] = useState(
+    [] as DetectorListItem[]
+  );
+  const [isLoadingFinalDetectors, setIsLoadingFinalDetectors] = useState<
+    boolean
+  >(true);
 
   // Getting all initial indices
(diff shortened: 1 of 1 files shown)


public/pages/DetectorsList/List/List.tsx
```diff
@@ -138,8 +146,16 @@ export const DetectorList = (props: ListProps) => {
...
 
-    dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+    const getUpdatedDetectors = async () => {
+      try {
+        await dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+      } catch (error) {
+        console.log('Error is found during getting detector list', error);
+        setIsLoadingFinalDetectors(false);
+      }
+    };
```
ohltyler, line 156: This is the only change; the other changes are auto-formatting changes. Somehow my local dev environment stopped auto-formatting on save at some point, fixed it now.

public/pages/DetectorsList/List/List.tsx
```diff
@@ -138,8 +146,16 @@ export const DetectorList = (props: ListProps) => {
       search: queryString.stringify(updatedParams),
     });
 
-    dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+    const getUpdatedDetectors = async () => {
+      try {
+        await dispatch(getDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS));
+      } catch (error) {
+        console.log('Error is found during getting detector list', error);
+        setIsLoadingFinalDetectors(false);
```
yizheliu-amazon, line 154: should we move`setIsLoadingFinalDetectors(false);` to `finally`? looks like only when exception is caught, `isLoading` can be set to false

ohltyler: That's the behavior I want here. It should only be set to false if there was an error. If successful and the detectors are retrieved, it should be set to true until the final set of sorted detectors to display are finished [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/DetectorsList/List/List.tsx#L169).

Note that `isLoadingFinalDetectors` is only related to sorting/filtering in the frontend. `isRequestingFromES` handles loading from the backend.

yizheliu-amazon: Got it.

Issue #192, opened 2020-06-02
Add start/stop batch actions on detector list page


ohltyler:
Add the ability to select a subset of displayed detectors on the detector list page and start/stop the detector jobs. Add a confirmation modal for each action, displaying to the user the affected detectors should they perform the action.

PR #195, opened 2020-06-03
Add start/stop batch actions on detector list page

ohltyler:
*Issue #, if available:* #192 

*Description of changes:*

This PR adds start/stop batch actions to the detector list page. 
A confirmation modal will pop up before any action is performed, informing the user of the detectors that will be affected should they decide to perform the action.

Specific additions & changes:
- Adds 'Actions' dropdown at top of page (with 'Start' and 'Stop' actions)
- Makes dropdown unavailable until a user has selected 1+ detectors
- Changes detector list to be selectable, where a user can select a subset of detectors
- Renamed ListControls -> ListFilters
- Will display toast notification if user tries to perform an action that will not affect any detectors
- Will display confirmation modals before performing an action
- Will display toast notification on status of action after performed (success/failure)
- Adds UT for all new files

Start detectors screenshot:
Screen Shot 2020-06-02 at 10 41 19 AM

Stop detectors screenshot:
Screen Shot 2020-06-02 at 10 41 54 AM

**Additional notes**
- Design and wording approved by UX
- All UT pass


public/pages/DetectorDetail/components/ConfirmModal/ConfirmModal.tsx
```diff
@@ -44,7 +44,7 @@ export const ConfirmModal = (props: ConfirmModalProps) => {
   return (
     <EuiModal onClose={props.onClose}>
       <EuiModalHeader>
-        <EuiModalHeaderTitle>{props.title}</EuiModalHeaderTitle>
+        <EuiModalHeaderTitle>{props.title}&nbsp;</EuiModalHeaderTitle>
```
ylwu-amzn, line 47: Why add `&nbsp;` here?

public/pages/DetectorsList/List/List.tsx
```diff
@@ -276,13 +306,131 @@ export const DetectorList = (props: ListProps) => {
...
+    });
+    await Promise.all(promises)
+      .then(() => {
+        toastNotifications.addSuccess(
+          'All selected detectors have been started successfully'
+        );
+      })
+      .catch(error => {
+        toastNotifications.addDanger(
+          `Error starting all selected detectors: ${error}`
```
ylwu-amzn, line 380: If multiple detectors fail, will show multiple toast notifications? Can we aggregate all errors and show one toast notification?

public/pages/DetectorsList/List/List.tsx
```diff
@@ -85,10 +99,10 @@ interface ListState {
 export const DetectorList = (props: ListProps) => {
   const dispatch = useDispatch();
   const allDetectors = useSelector((state: AppState) => state.ad.detectorList);
+  const allMonitors = useSelector((state: AppState) => state.alerting.monitors);
```
ylwu-amzn, line 102: Will this cause performance issue? Default monitor count is 1000, but user can change it. How about we limit how many detectors can be deleted in a batch call and get their related monitors when click start/stop button?

public/pages/DetectorDetail/components/ConfirmModal/ConfirmModal.tsx
ohltyler: Spacing issue where the last character (in this case a `?`) gets cut off by the close icon. This adds an extra space to always prevent this from happening.

Screen Shot 2020-06-02 at 6 23 16 PM

public/pages/DetectorsList/List/List.tsx
ohltyler: `Promise.all()` tries to resolve all of the promises. If any fails, it aborts everything and throws an exception. So this will only show once.

public/pages/DetectorsList/List/List.tsx
ohltyler: Adding selection to the table limits the maximum number of selected detectors to be whatever maximum page size there is (in this case, 50). Selection can't span across multiple table pages. So, the batch limit here is 50. I've tested starting/stopping 50 detectors at once and it seems to not have major performance loss.

public/pages/DetectorsList/List/List.tsx
ohltyler: The same thing applies to the success toast notification; it will only be shown once after a batch action is performed.

public/pages/DetectorsList/utils/helpers.ts
```diff
@@ -56,3 +58,33 @@ export const getDetectorStateOptions = () => {
     text: detectorState,
   }));
 };
+
+export const getValidDetectors = (
```
yizheliu-amazon, line 62: maybe you can name it like `getDetectorsForAction`

public/pages/DetectorsList/List/List.tsx
yizheliu-amazon: will it only show error for 1st failed detector, even there are multiple failed detectors?

ohltyler: > will it only show error for 1st failed detector, even there are multiple failed detectors?

Kind of. Because `Promise.all()` fails out once a single failure has occurred, the rest of the promises won't be resolved, so technically only one detector could fail. But yes, the error message here would be related to the first failed detector

public/pages/DetectorsList/utils/helpers.ts
ohltyler: that's a better name, will change

Issue #206, opened 2020-06-08
Change UT workflow to run against new pull requests


ohltyler:
Change the workflow to use the open source Kibana repository so we can run the unit tests workflow when PRs are made with forked repositories.

ohltyler:
Closing. No clear timeline on when this will be possible to do.

PR #207, opened 2020-06-08
Change UT workflow to run against PRs

ohltyler:
*Issue #, if available:* #206 

*Description of changes:*

This PR changes the UT workflow by adding functionality to run against new pull requests made to the `master` branch

### DIFF
--- a/.github/workflows/unit-tests-workflow.yml
+++ b/.github/workflows/unit-tests-workflow.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   tests:
@@ -15,9 +18,8 @@ jobs:
       - name: Checkout Kibana
         uses: actions/checkout@v2
         with:
-          repository: opendistro-for-elasticsearch/kibana-oss
-          ref: 7.7.0
-          token: ${{ secrets.KIBANA_OSS_ACCESS }}
+          repository: elastic/kibana
+          ref: v7.7.0
           path: kibana
       - name: Get node and yarn versions
         id: versions_step


.github/workflows/unit-tests-workflow.yml
```diff
@@ -15,9 +18,8 @@ jobs:
       - name: Checkout Kibana
         uses: actions/checkout@v2
         with:
-          repository: opendistro-for-elasticsearch/kibana-oss
-          ref: 7.7.0
-          token: ${{ secrets.KIBANA_OSS_ACCESS }}
+          repository: elastic/kibana
```
ylwu-amzn, line 21: Is it possible we passed `elastic/kibana` but fail `opendistro-for-elasticsearch/kibana-oss` ? As we are going to release plugin on ODFE, suggest to run test workflow on `opendistro-for-elasticsearch/kibana-oss`.

Have you tried `provide an action with a secret as an input or environment variable`?

yizheliu-amazon: The root cause is correct. However, if contributors outside of our team want to contribute to our product, they will be blocked by such secrets.

I think the reason we use `elastic/kibana` is that our product(AD Kibana plugin) is open-sourced, it is expected to be compatible with OSS Kibana, instead of a private repo, like `opendistro-for-elasticsearch/kibana-oss`. We open-source AD in order to get attention from external contributors, and they must be using `elastic/kibana` or any other public oss kibana.

If it can go public, we can just use public  `opendistro-for-elasticsearch/kibana-oss`, but I think it may involve more legal discussion. We can propose this idea to leadership. 

 If our UT passes for `elastic/kibana`, but fail for `opendistro-for-elasticsearch/kibana-oss`, I would say we should figure out root cause and decide which solution to go. In that case, there must be some breaking changes between `elastic/kibana` and `opendistro-for-elasticsearch/kibana-oss` even at same version, it is worth the investigation.

ohltyler: As stated in the link in the PR description, the problem is no secrets can be passed to the github runner if it is coming from a forked branch. There is a security vulnerability if that was not the case. This is why originally this workflow was only running after **pushes** to master rather than **pull requests from forked branches* to master. Me and @yizheliu-amazon were thinking of adding the ability to test against PRs by running our plugin using the open source Kibana repo, rather than the private one which requires a secret.

ylwu-amzn: Sure, have pinged Alolita about whether we should use elastic/kibana for testing workkfow and whether we can make `opendistro-for-elasticsearch/kibana-oss ` public.

Another work around may be we use `elastic/kibana` for PR triggered test workflow, and `opendistro-for-elasticsearch/kibana-oss` for push triggered test workflow. So we can test both.

ohltyler: That could also work. Will hold off on merging until you hear back. Thanks for checking

ohltyler:
Will close this PR for now. Once a decision is made, will make changes to all workflows in one PR (CD action, UT action)

Issue #75, opened 2020-05-01
Unit test workflow failing when running yarn kbn bootstrap


ohltyler:
When looking at the logs of the failed GitHub actions, after running `yarn kbn bootstrap`:

`error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/home/runner/work/anomaly-detection-kibana-plugin/anomaly-detection-kibana-plugin/kibana/packages/kbn-plugin-helpers/node_modules/which'".`

I've had this happen to me before locally, seemed to be fixed when running `yarn kbn bootstrap` a second time. Need to investigate

ohltyler:
Changes have been addressed to remediate the issue for now. See https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/104. Will remove bug label.

ohltyler:
After revisiting, I may have fixed the issue. I ran [yarn upgrade](https://classic.yarnpkg.com/en/docs/cli/upgrade/) to update all dependencies. Seems to not run into errors as mentioned above. Tested out workflow that only runs `yarn kbn bootstrap` once, is passing now. See successful workflow [here](https://github.com/ohltyler/anomaly-detection-kibana-plugin/runs/754868897?check_suite_focus=true)

ylwu-amzn:
Cool, that's great.

PR #210, opened 2020-06-09
Fix UT workflow to only run bootstrap once

ohltyler:
*Issue #, if available:* #75

*Description of changes:*

This PR fixes the bug where running `yarn kbn bootstrap` would occasionally fail the first time and would have to be ran twice (on both the GitHub runners and on a local dev environment).

Specific changes:
- Removed the extra bootstrap step in the UT workflow
- Upgraded the `yarn.lock` by running `yarn upgrade` to upgrade all dependencies. This fixes the bootstrap failing issue.
- Updated test snapshots after upgrading dependencies.

Notes
- Confirmed all UT pass
- Confirmed UT workflow passes (see [this run](https://github.com/ohltyler/anomaly-detection-kibana-plugin/actions/runs/130088218))
- Confirmed the plugin loads and functions fine after upgrading

### DIFF
--- a/.github/workflows/unit-tests-workflow.yml
+++ b/.github/workflows/unit-tests-workflow.yml
@@ -2,7 +2,7 @@ name: Unit tests workflow
 on:
   push:
     branches:
-      - fixBootstrap
+      - master
 
 jobs:
   tests:
--- a/.github/workflows/unit-tests-workflow.yml
+++ b/.github/workflows/unit-tests-workflow.yml
@@ -2,7 +2,7 @@ name: Unit tests workflow
 on:
   push:
     branches:
-      - master
+      - fixBootstrap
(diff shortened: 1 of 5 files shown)


yizheliu-amazon:
can you quickly do small smoke test to verify if UI is working without anything broken? looks like many dependencies are upgraded.

ohltyler:
> can you quickly do small smoke test to verify if UI is working without anything broken? looks like many dependencies are upgraded.

yes, I've done this. don't see anything broken

yizheliu-amazon:
> > can you quickly do small smoke test to verify if UI is working without anything broken? looks like many dependencies are upgraded.
> 
> yes, I've done this. don't see anything broken

Cool.

PR #221, opened 2020-06-12
Add more e2e test cases for Dashboard/Detector list

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Add more e2e test cases for Dashboard/Detector list. Next PR will set up code coverage.


.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
```diff
@@ -13,31 +13,67 @@
...
-  it('Empty dashboard - no detector', () => {
-    cy.server();
-    cy.route(
-      'GET',
-      [API_URL_PREFIX, AD_PATH, DETECTORS + '*'].join(SLASH),
-      'fixture:no_detector_index_response.json'
-    ).as('getDetectors');
+  it('Empty dashboard - no detector index', () => {
+    cy.mockGetDetectorOnAction('no_detector_index_response.json', () => {
+      cy.visit([APP_URL_PREFIX, AD_URL, DASHBOARD].join(SLASH));
```
ohltyler, line 26: I wonder if we should add constants for the page url at the top of the file, so we don't have to repeatedly write `[APP_URL_PREFIX, AD_URL, DASHBOARD].join(SLASH)` all the time

ohltyler: In fact, the same thing comes up in the `commands` file. Maybe move all of these to some url constants somewhere? Then we would have less constants to import as well.

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
```diff
@@ -13,31 +13,67 @@
...
+  });
+
+  it('AD dashboard - create detector', () => {
+    cy.mockGetDetectorOnAction('no_detector_index_response.json', () => {
+      cy.visit([APP_URL_PREFIX, AD_URL, DASHBOARD].join(SLASH));
+    });
+
+    cy.mockSearchIndexOnAction('search_index_response.json', () => {
+      cy.get('.euiButton--primary.euiButton--fill:first').click({
+        force: true,
```
ohltyler, line 57: why do we have to set `force: true` for all click/type actions?

.cypress/fixtures/single_detector_index_response.json
```diff
@@ -0,0 +1,43 @@
+{
```
ohltyler, line 1: should this be renamed to something like `single_detector_response` rather than saying `index` here? seems confusing to mention `index`

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
```diff
@@ -13,31 +13,67 @@
...
 
-    cy.contains('h2', 'You have no detectors');
+    cy.contains('h3', 'Live anomalies');
+    cy.contains(
+      'p',
+      'All matching detectors are under initialization or stopped for the last 30 minutes. Please adjust filters or come back later.'
+    );
+  });
+
+  it('AD dashboard - create detector', () => {
```
ohltyler, line 50: should we move this test to a suite pertaining to the create detector page? and maybe just simplify this test to just make sure that when clicking on the create detector button we go to the create detector page?

yizheliu-amazon: Good point. moved to a separate suite for create detector.

.cypress/fixtures/single_detector_index_response.json
yizheliu-amazon: sure. renamed it to `single_detector_response`

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
yizheliu-amazon: Kibana body has a <div.content> which has `overflow: hidden`, and it is the parent of all divs in the page. As per [cypress doc](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Visibility), such element will be considered as invisible. If `force` is not set as true, error will show up as below:

Screen Shot 2020-06-16 at 1 03 47 PM

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
yizheliu-amazon: good point. use separate method to get URLs instead.

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
```diff
@@ -13,31 +13,47 @@
...
+    );
+  });
+
+  it('AD dashboard - redirect to create detector', () => {
+    cy.mockGetDetectorOnAction('no_detector_index_response.json', () => {
+      cy.visit(buildAdAppUrl(DASHBOARD));
+    });
+
+    cy.mockSearchIndexOnAction('search_index_response.json', () => {
+      cy.get('.euiButton--primary.euiButton--fill:first').click({
```
ylwu-amzn, line 52: Can we use a simpler way to get the button? Like via name?

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
ohltyler: Got it, thanks

.cypress/integration/ad/dashboard/ad_dashboard.spec.ts
yizheliu-amazon: 'Create detector' text is inside `<span>`, which is inside the button element `<a>`
Screen Shot 2020-06-16 at 2 07 20 PM

Getting it via name only returns the `<span>`, which is not clickable. 

The simpler way is to use `span[data-test-subj="add_detector"]`

Issue #223, opened 2020-06-12
Improve release automation in CD action


ohltyler:
Currently, the CD action builds and pushes a built zip to s3. Add more automatic functionality to the CD, such as creating a draft release, adding built zip to that draft release, etc.

PR #226, opened 2020-06-16
Automate release notes generation on pushes to master

ohltyler:
*Issue #, if available:* #223 

*Description of changes:*

This PR adds an action ([details here](https://github.com/release-drafter/release-drafter)) to automatically create and update draft release notes by adding new merged PR commits into the notes in an organized fashion.

This also removes the unnecessary second bootstrap command in the CD action, which was fixed by #210 

**NOTE** this will require new PRs to be labeled so they can be correctly categorized in the release notes. For example, if a PR is considered an `enhancement`, it will need to be labeled with the `enhancement` tag. Then, when merged into master, the release notes will add the commit under the `Enhancements` header automatically.

Example (on my forked repo): 
Screen Shot 2020-06-16 at 11 47 31 AM

### DIFF
--- a/.github/draft-release-notes-config.yml
+++ b/.github/draft-release-notes-config.yml
@@ -20,6 +20,7 @@ categories:
   - title: 'Bug Fixes'
     labels:
       - 'bug'
+      - 'bug fix'
   - title: 'Infra Changes'
     labels:
       - 'infra'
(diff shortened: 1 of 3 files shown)


yizheliu-amazon:
This is cool!! Thanks. If PR is not labeled with tag, what category will it be considered as in release note?

ohltyler:
> This is cool!! Thanks. If PR is not labeled with tag, what category will it be considered as in release note?

It will still be included, but will just be unlabeled at the top of the draft. In fact, an example of an unlabeled PR is in the screenshot in the description (the update workflows and release notes dummy PR)

Issue #231, opened 2020-06-20
Add more Cypress tests for dashboard and detector list pages


ohltyler:
Add test cases for the home dashboard and detector list pages with Cypress to help increase code coverage.

PR #234, opened 2020-06-22
Add Cypress tests for dashboard and detector list

ohltyler:
*Issue #, if available:* #231 

*Description of changes:*

This PR adds several Cypress integration tests to test various workflows on the 2 home pages of the plugin (dashboard and detector list).

Dashboard:
- tests filtering of the dropdown filter

Detector list:
- tests filtering of the dropdown filter
- tests rendering of the page when different indices are empty
- tests batch action functionality (start/stop/delete)

### DIFF
--- a/.cypress/support/commands.ts
+++ b/.cypress/support/commands.ts
@@ -12,7 +12,15 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-import { DETECTORS, INDICES_PATH, MAPPINGS_PATH } from '../utils/constants';
+import {
+  DETECTORS,
+  INDICES_PATH,
+  MAPPINGS_PATH,
+  START_PATH,
+  STOP_PATH,
+  SLASH,
+  SEARCH_PATH,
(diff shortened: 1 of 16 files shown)


.cypress/support/commands.ts
```diff
@@ -63,6 +71,20 @@ Cypress.Commands.add('mockSearchIndexOnAction', function(
   cy.wait('@getIndices');
 });
 
+Cypress.Commands.add('mockSearchOnAction', function(
```
yizheliu-amazon, line 74: looks like this is not used yet. Maybe rename to `mockSearchDetectorOnAction`?

ohltyler: good catch, I was going to use it for some dashboard tests but ended up not needing it yet. I was thinking to leave it as `mockSearchOnAction` since this can be used for searching anything in ES, similar to the `searchES()` function.

yizheliu-amazon: Sure. sounds good to me.

Issue #215, opened 2020-06-10
Can I choose remote cluster indices for anomaly detection?


HagunKim:
I'd like to use anomaly-detection plugin for other ES cluster connected remote cluster configuration.
In Kibana, I can use discover or something using search with the remote cluster(cross cluster search), but anomaly-detection doesn't support it.
Is it possible to use cross cluster indices for anomaly detection in kibana?

ylwu-amzn:
> I'd like to use anomaly-detection plugin for other ES cluster connected remote cluster configuration.
> In Kibana, I can use discover or something using search with the remote cluster(cross cluster search), but anomaly-detection doesn't support it.
> Is it possible to use cross cluster indices for anomaly detection in kibana?

Can you explain more about what you want for `use cross cluster indices for anomaly detection in kibana`?

As kaituo said, we don't support this feature in kibana now. Can you try [create detector API](https://opendistro.github.io/for-elasticsearch-docs/docs/ad/api/#create-anomaly-detector) ?

HagunKim:
@kaituo, @ylwu-amzn, Thanks for your reply! For now, when I try to create detector on Kibana, In dropdown list for index, I can only choose indices of ES that is connected directly.
I want to see and choose my Kibana index patterns containing cross cluster indices or the cross cluster indices themselves on the dropdown list.
Could you consider this modifications?

ylwu-amzn:
Thanks @HagunKim, sure, that's good suggestion. Have you tried [create detector API](https://opendistro.github.io/for-elasticsearch-docs/docs/ad/api/#create-anomaly-detector)?

HagunKim:
@ylwu-amzn , I tried create detector API with remote cluster indices via Kiabana Dev Tools menu. And it worked. I got 201 response and I saw the detector on Kibana. When I tried to create using non-existing index, it returned error with 404 response.

ylwu-amzn:
> @ylwu-amzn , I tried create detector API with remote cluster indices via Kiabana Dev Tools menu. And it worked. I got 201 response and I saw the detector on Kibana. When I tried to create using non-existing index, it returned error with 404 response.

Cool, thanks @HagunKim. I created a feature issue to track this requirement https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/220,  Feel free to add your suggestion there. 

If no more questions, how about we close this question issue?

HagunKim:
[...]

PR #244, opened 2020-06-25
support remote cluster indices

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
Community has requirements to create detector on top of remote cluster’s indices, refer to Github issue (https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/215). User can do this with create detector public API, but can’t choose/input remote cluster indices and its fields on Kibana.

Currently Elasticsearch doesn’t support listing indices on remote cluster, nor get index mapping.  So we just support remote cluster indices by allowing user input index and filed name manually. 

## Create/edit detector page
Screen Shot 2020-06-24 at 10 12 59 PM

## Edit feature page
Screen Shot 2020-06-24 at 10 12 03 PM

### DIFF
--- a/public/pages/createDetector/containers/DataSource/DataSource.tsx
+++ b/public/pages/createDetector/containers/DataSource/DataSource.tsx
@@ -93,7 +93,7 @@ function DataSource(props: DataFilterProps) {
       {isRemoteIndex() ? (
         <div>
           <EuiCallOut
-            title="This detector is using remote cluster index. Please input time field and filter field manually."
+            title="This detector is using a remote cluster index, so you need to manually input the time field."
             color="warning"
             iconType="alert"
           />
--- a/public/pages/createDetector/containers/DataSource/DataSource.tsx
+++ b/public/pages/createDetector/containers/DataSource/DataSource.tsx
@@ -13,9 +13,9 @@
  * permissions and limitations under the License.
  */
 
-import { EuiComboBox, EuiSelect } from '@elastic/eui';
+import { EuiComboBox, EuiCallOut, EuiSpacer } from '@elastic/eui';
 import { Field, FieldProps } from 'formik';
-import { debounce, get } from 'lodash';
+import { debounce, get, isEmpty } from 'lodash';
 import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { CatIndex, IndexAlias } from '../../../../../server/models/types';
@@ -39,6 +39,7 @@ import { FormattedFormRow } from '../../components/FormattedFormRow/FormattedFor
 function DataSource(props: DataFilterProps) {
   const dispatch = useDispatch();
   const [queryText, setQueryText] = useState('');
+  const [indexName, setIndexName] = useState(undefined);
   const elasticsearchState = useSelector(
     (state: AppState) => state.elasticsearch
   );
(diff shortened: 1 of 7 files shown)


ohltyler:
Do we want to get UX input on any of the changes and wording?

public/redux/reducers/elasticsearch.ts
```diff
@@ -93,7 +94,7 @@ const reducer = handleActions<ElasticsearchState>(
       ): ElasticsearchState => ({
         ...state,
         requesting: false,
-        errorMessage: action.error.data.error,
+        errorMessage: get(action, 'error.data.error', action.error),
```
ohltyler, line 97: this looks like this would return `action.error.data.error` if exists, or `action.error` if not. Wouldn't it always be one or the other?

public/pages/createDetector/containers/DataSource/DataSource.tsx
```diff
@@ -131,11 +155,25 @@ function DataSource(props: DataFilterProps) {
...
+              }}
+              onBlur={() => {
+                form.setFieldTouched('timeField', true);
+              }}
+              onChange={options => {
+                form.setFieldValue('timeField', get(options, '0.label'));
+              }}
+              selectedOptions={(field.value && [{ label: field.value }]) || []}
+              singleSelection={true}
+              isClearable={false}
```
ohltyler, line 176: Could we maybe change this to `isClearable={true}`? I've noticed it seems kind of annoying when trying to change the selected index in the combobox. The only way to do it is backspace while the cursor is in the box.

public/redux/reducers/elasticsearch.ts
ylwu-amzn: Not sure about it, haven't done all error cases. To keep compatible with original cases and handle remote indices error, it's safe to fall back to action.error as action is not undefined. There are other places where error handling use different ways, created an issue to track this problem and fix it later.https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/245

ylwu-amzn:
> Do we want to get UX input on any of the changes and wording?

Ping UX and tech writer. Waiting for reply.

ylwu-amzn:
> > Do we want to get UX input on any of the changes and wording?
> 
> Ping UX and tech writer. Waiting for reply.

Add another callout based on UX designer's suggestion: "consider diaplaying something within the Add filter section as the user may scroll past the callout’s visibility."

Screen Shot 2020-06-25 at 12 22 09 PM

ohltyler:
Are the first 2 screenshots out of date, in that the wording has changed to match the recent screenshots?

public/pages/createDetector/containers/DataSource/DataSource.tsx
```diff
@@ -131,11 +155,25 @@ function DataSource(props: DataFilterProps) {
...
+            <EuiComboBox
               id="timeField"
               placeholder="Find timestamp"
               options={timeStampFieldOptions}
+              onSearchChange={handleSearchChange}
+              onCreateOption={(createdOption: string) => {
+                const normalizedOptions = createdOption.trim();
+                if (!normalizedOptions) return;
+                form.setFieldValue('timeField', normalizedOptions);
+              }}
```
yizheliu-amazon, line 167: why allow user to create option? what if user inputs invalid timestamp field?

ylwu-amzn: Similar questions: what if user input invalid filed in custom expression of filter and feature definition.
If user input invalid filed, like they input a filed which is string not timestamp, or the filed not exists, the detector just can't get any data to run. 
A better way is we can pop up some warning message and block create detector with invalid parameters. 
Amit is working on auto AD project. We can discuss that design, a common sanity checker component is useful. And I think that can be a part of Amit's project. Create a backlog issue to track this.

ylwu-amzn: > why allow user to create option?

This change is not a final solution, just some workaround to support remote cluster indices. 
For local indices, user still can select filed from dropdown list. For remote cluster indices, ES doesn't support listing remote cluster's indices, not get index mapping. We can't get remote cluster index's fields. So for remote cluster indices, user need to input filed manually.

yizheliu-amazon: Thanks. Btw, why timestamp field from remote index is not visible?

ylwu-amzn: > Thanks. Btw, why timestamp field from remote index is not visible?

For local index, we can get its index mapping and parse which filed is timestamp type. But for remote cluster index, we can't get its index mapping. So we don't know which filed is timestamp.

PR #247, opened 2020-06-30
Fix default filter query when creating detector

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR fixes the default custom filter query. Before, there was an extra `query` field in the default JSON, which was throwing errors when trying to create the detector. This removes that extra field.

Before:
Screen Shot 2020-06-30 at 9 22 03 AM

After:
Screen Shot 2020-06-30 at 9 21 26 AM

Notes:
- all UT pass
- confirmed custom filter functionality works


kaituo:
What error does this throw?

ohltyler:
> What error does this throw?

Screen Shot 2020-06-30 at 8 41 33 AM

PR #248, opened 2020-07-07
Add missing feature alert if recent feature data is missing

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Add missing feature alert if recent feature data is missing

If latest 2 feature data points of any feature are missing, we show yellow warning like below
Screen Shot 2020-07-06 at 10 22 20 PM

If latest 3 feature data points of any feature are missing, we show red alert below:
Screen Shot 2020-07-07 at 10 21 11 AM

Regardless of yellow or red alert, as long as there exists missing feature data, there is annotation on the feature chart. 

Screen Shot 2020-07-06 at 10 25 17 PM

_Meanwhile, I will check with tech writer on wording change_ >> Done

When data point is missing, below is shown:
Screen Shot 2020-07-09 at 2 44 39 PM

### DIFF
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -117,20 +117,24 @@ export const FeatureChart = (props: FeatureChartProps) => {
   const featureData = prepareDataForChart(props.featureData, props.dateRange);
 
   // return undefined if featureMissingDataPointAnnotationStartDate is missing
+  // OR it is even behind the specified date range
   const getFeatureMissingAnnotationDateRange = (
     dateRange: DateRange,
     featureMissingDataPointAnnotationStartDate?: number
   ) => {
-    if (!featureMissingDataPointAnnotationStartDate) {
-      return undefined;
+    if (
+      featureMissingDataPointAnnotationStartDate &&
+      dateRange.endDate > featureMissingDataPointAnnotationStartDate
+    ) {
+      return {
+        startDate: Math.max(
+          dateRange.startDate,
+          featureMissingDataPointAnnotationStartDate
+        ),
+        endDate: dateRange.endDate,
+      };
     }
-    return {
-      startDate: Math.max(
-        dateRange.startDate,
-        featureMissingDataPointAnnotationStartDate
-      ),
-      endDate: dateRange.endDate,
-    };
+    return undefined;
   };
 
   return (
--- a/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
+++ b/public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
@@ -22,8 +22,10 @@ import {
   Position,
   Settings,
   ScaleType,
+  LineAnnotation,
+  AnnotationDomainTypes,
 } from '@elastic/charts';
-import { EuiEmptyPrompt, EuiText, EuiLink, EuiButton } from '@elastic/eui';
+import { EuiText, EuiLink, EuiButton, EuiIcon } from '@elastic/eui';
 import React, { useState, Fragment } from 'react';
 import ContentPanel from '../../../../components/ContentPanel/ContentPanel';
 import { useDelayedLoader } from '../../../../hooks/useDelayedLoader';
@@ -32,14 +34,15 @@ import {
   FeatureAttributes,
   DateRange,
   FEATURE_TYPE,
+  Schedule,
 } from '../../../../models/interfaces';
 import { darkModeEnabled } from '../../../../utils/kibanaUtils';
-import { prepareDataForChart } from '../../../utils/anomalyResultUtils';
-import { CodeModal } from '../../../DetectorConfig/components/CodeModal/CodeModal';
 import {
-  CHART_FIELDS,
-  FEATURE_CHART_THEME,
-} from '../../utils/constants';
+  prepareDataForChart,
+  getFeatureMissingDataAnnotations,
+} from '../../../utils/anomalyResultUtils';
+import { CodeModal } from '../../../DetectorConfig/components/CodeModal/CodeModal';
+import { CHART_FIELDS, FEATURE_CHART_THEME } from '../../utils/constants';
 
 interface FeatureChartProps {
   feature: FeatureAttributes;
@@ -54,6 +57,10 @@ interface FeatureChartProps {
   featureDataSeriesName: string;
   edit?: boolean;
   onEdit?(): void;
+  detectorInterval: Schedule;
+  showFeatureMissingDataPointAnnotation?: boolean;
+  detectorEnabledTime?: number;
+  rawFeatureData: FeatureAggregationData[];
 }
 const getDisabledChartBackground = () =>
   darkModeEnabled() ? '#25262E' : '#F0F0F0';
@@ -108,6 +115,24 @@ export const FeatureChart = (props: FeatureChartProps) => {
   );
 
   const featureData = prepareDataForChart(props.featureData, props.dateRange);
+
+  // return undefined if featureMissingDataPointAnnotationStartDate is missing
+  const getFeatureMissingAnnotationDateRange = (
+    dateRange: DateRange,
+    featureMissingDataPointAnnotationStartDate?: number
+  ) => {
+    if (!featureMissingDataPointAnnotationStartDate) {
+      return undefined;
+    }
+    return {
+      startDate: Math.max(
+        dateRange.startDate,
+        featureMissingDataPointAnnotationStartDate
+      ),
+      endDate: dateRange.endDate,
+    };
+  };
+
   return (
     <ContentPanel
       title={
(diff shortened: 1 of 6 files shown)


ylwu-amzn:
> If latest 2 feature data points of any feature are missing, we show yellow warning

Why don't show yellow warning if 1 feature data point missing? So user can know the data ingestion may have problem earlier.

yizheliu-amazon:
> > If latest 2 feature data points of any feature are missing, we show yellow warning
> 
> Why don't show yellow warning if 1 feature data point missing? So user can know the data ingestion may have problem earlier.

It may be too sensitive if latest 1 data point is missing, and ingestion is recovered soon. This may happen if data ingestion is delayed occasionally. If latest 2 consecutive data points are missing, it can be considered as consistent behavior and we should let user be aware of this.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,274 @@ export const parsePureAnomalies = (
...
+  });
+  return featureDataPointsForDetector;
+};
+
+export const getFeatureMissingSeverities = (featuresDataPoint: {
+  [key: string]: FeatureDataPoint[];
+}): Map<MISSING_FEATURE_DATA_SEVERITY, string[]> => {
+  const featureMissingSeverities = new Map();
+
+  for (const [featureName, featureDatePoints] of Object.entries(
```
ohltyler, line 714: small: typo in `featureDataPoints`

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,274 @@ export const parsePureAnomalies = (
...
+};
+
+export const getFeatureDataMissingMessageAndActionItem = (
+  featureMissingSev: MISSING_FEATURE_DATA_SEVERITY | undefined,
+  featuresWithMissingData: string[]
+) => {
+  switch (featureMissingSev) {
+    case MISSING_FEATURE_DATA_SEVERITY.YELLOW:
+      return {
+        message: `Recent data is missing for feature ${featuresWithMissingData.join(
```
ohltyler, line 777: if > 1 feature with missing data, should probably say "Recent data is missing for features" rather than "feature"

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,274 @@ export const parsePureAnomalies = (
...
+          ', '
+        )}, anomaly result is missing at the same time because of that.`,
+        actionItem:
+          'Make sure your data is ingested correctly. See the feature data shown below for more details.',
+      };
+    case MISSING_FEATURE_DATA_SEVERITY.RED:
+      return {
+        message: `Data ingestion is not going well for feature ${featuresWithMissingData.join(
```
ohltyler, line 785: same thing about "feature" v. "features" here. I'm guessing UX will go over this as well

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -84,21 +89,57 @@ export const getAnomalyResultsWithDateRange = (
...
+) => {
+  const params = buildParamsForGetAnomalyResultsWithDateRange(
+    startTime,
+    endTime
+  );
+  dispatch(getDetectorResults(detectorId, params));
 };
 
 const MAX_DATA_POINTS = 1000;
+const MAX_FEATURE_ANNOTATIONS = 100;
```
ohltyler, line 108: why not have `MAX_FEATURE_ANNOTATIONS` be the same amount as `MAX_DATA_POINTS`?

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,274 @@ export const parsePureAnomalies = (
...
+};
+
+const getDataPointsInWindow = (
+  dataPoints: FeatureDataPoint[],
+  timeWindow: DateRange
+) => {
+  return dataPoints.filter(
+    dataPoint =>
+      get(dataPoint, 'plotTime', 0) >= timeWindow.startDate &&
+      get(dataPoint, 'plotTime', 0) < timeWindow.endDate
```
ohltyler, line 624: is there a reason you filter out by just `<` end date here, and filter out by `<=` end date in other places?

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: Thanks for the catch.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: `feature` here is more like adjective, because feature name(s) is shown after `feature`. I will confirm this with tech writer.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: sure. I will confirm this with tech writer or UX.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: 1000 annotations are too crowded on chart. Using 100 looks good

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: that's good catch. using `<=` is a bug which I will fix in next commit.

public/pages/utils/anomalyResultUtils.ts
ohltyler: If there's any missing feature data points, we want to show this missing feature annotation alert, right? So if there's >100 missing points, wouldn't we miss some annotations that should be there?

yizheliu-amazon: Yes. Extra missing points won't be shown in that case. Instead, I add sampling on all the missing points, and annotations are generated from those samples. Those annotations can cover all the missing points

ohltyler: Got it. Makes more sense looking through the code again. Thanks!

yizheliu-amazon:
Issue associated with this: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/249

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
```diff
@@ -64,7 +64,7 @@ export const FeatureBreakDown = React.memo((props: FeatureBreakDownProps) => {
       props.detector.enabledTime &&
       props.isFeatureDataMissing ? (
         <EuiCallOut
-          title={`Show missing data only after last enable time: ${moment(
+          title={`Show missing data only after last enabled time: ${moment(
```
ohltyler, line 67: Wording seems kind of confusing here, maybe something like 'Missing data is only shown since the detector was last enabled on <enabled-time>', or 'There is missing data since the detector was last enabled on <enabled-time>'?

ohltyler: Or even 'There is missing data since the detector was last enabled on (last enabled time)'

ohltyler:
Also, general question, is there an easy way to remove the line connecting the feature data points where data was missing in between? Feel like it shouldn't be included.
Screen Shot 2020-07-09 at 7 03 07 PM

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
yizheliu-amazon: Thanks for the idea. Maybe use `Missing data is only shown since last enabled time: <time-string>`?

yizheliu-amazon: just want to keep wording concise.

ohltyler: > Thanks for the idea. Maybe use `Missing data is only shown since last enabled time: <time-string>`?

Sounds good to me, agreed to keep it short

yizheliu-amazon:
> Also, general question, is there an easy way to remove the line connecting the feature data points where data was missing in between? Feel like it shouldn't be included.
> Screen Shot 2020-07-09 at 7 03 07 PM

Actually this is a corner case. My local ES process somehow stopped itself during that period. Ideally it should not happen. And the connecting line is automatically drawn by Elastic chart library. Let me explore if it is possible to skip it.

yizheliu-amazon:
> > Also, general question, is there an easy way to remove the line connecting the feature data points where data was missing in between? Feel like it shouldn't be included.
> > Screen Shot 2020-07-09 at 7 03 07 PM
> 
> Actually this is a corner case. My local ES process somehow stopped itself during that period. Ideally it should not happen. And the connecting line is automatically drawn by Elastic chart library. Let me explore if it is possible to skip it.

After exploring, I didn't find good way to remove the connecting line. Issue created: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/250

ohltyler:
> > > Also, general question, is there an easy way to remove the line connecting the feature data points where data was missing in between? Feel like it shouldn't be included.
> > > Screen Shot 2020-07-09 at 7 03 07 PM
> > 
> > 
> > Actually this is a corner case. My local ES process somehow stopped itself during that period. Ideally it should not happen. And the connecting line is automatically drawn by Elastic chart library. Let me explore if it is possible to skip it.
> 
> After exploring, I didn't find good way to remove the connecting line. Issue created: #250

Yeah, not a big deal, thanks for looking into it!

public/utils/constants.ts
```diff
@@ -65,3 +65,14 @@ export const NAME_REGEX = RegExp('^[a-zA-Z0-9._-]+$');
...
 //https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java#L186
 export const SHINGLE_SIZE = 8;
+
+export const FEATURE_DATA_POINTS_WINDOW = 3;
+
+export enum MISSING_FEATURE_DATA_SEVERITY {
+  // user attention not needed
+  GREEN = '0',
+  // needs user attention
+  YELLOW = '1',
```
ylwu-amzn, line 75: Will show yellow warning if only 1 feature data point missing?

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
```diff
@@ -125,72 +154,86 @@ export const FeatureChart = (props: FeatureChartProps) => {
...
+                  id="featureMissingAnnotations"
+                  domainType={AnnotationDomainTypes.XDomain}
+                  dataValues={getFeatureMissingDataAnnotations(
+                    props.showFeatureMissingDataPointAnnotation
+                      ? props.rawFeatureData
+                      : props.featureData,
+                    props.detectorInterval.interval,
+                    getFeatureMissingAnnotationDateRange(
+                      props.dateRange,
+                      props.detectorEnabledTime
```
ylwu-amzn, line 199: So we will only show missing data alerts from latest detector enabled time? Is it possible we show alerts for data before?

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -84,21 +89,57 @@ export const getAnomalyResultsWithDateRange = (
...
+const MAX_FEATURE_ANNOTATIONS = 100;
 
 const calculateStep = (total: number): number => {
   return Math.ceil(total / MAX_DATA_POINTS);
 };
 
-// If array size is 100K, `findAnomalyWithMaxAnomalyGrade` 
-// takes less than 2ms by average, while `Array#reduce` 
-// takes about 16ms by average and`Array#sort` 
+const calculateSampleWindowsWithMaxDataPoints = (
```
ylwu-amzn, line 114: Feature data points are already sampled, should we sample again for the missing data points? Can we show missing alert annotation for every feature data points.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,288 @@ export const parsePureAnomalies = (
...
+    // skip checking for latest interval as data point for it may not be delivered in time
+    getRoundedTimeInMin(
+      dateRange.endDate -
+        FEATURE_DATA_CHECK_WINDOW_OFFSET * interval * MIN_IN_MILLI_SECS
+    );
+    currentTime += interval * MIN_IN_MILLI_SECS
+  ) {
+    const isExisting = findTimeExistsInWindow(
+      existingTimes,
+      getRoundedTimeInMin(currentTime),
```
ylwu-amzn, line 563: Why get rounded time in minute?  Are the `existingTimes` rounded time?

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -483,3 +523,288 @@ export const parsePureAnomalies = (
...
+  if (!dateRange) {
+    return featureDataPoints;
+  }
+
+  const existingTimes = isEmpty(featureData)
+    ? []
+    : featureData
+        .map(feature => getFloorPlotTime(feature.startTime))
+        .filter(featureTime => featureTime != undefined);
+  for (
```
ylwu-amzn, line 551: The time complexity of this block is O(m*n). m is intervals, n is existingTimes count. Have you tested the performance? Suggest to use binary search in function `findTimeExistsInWindow` as existingTimes is ordered.

public/pages/AnomalyCharts/components/FeatureChart/FeatureChart.tsx
yizheliu-amazon: Yes, only since latest enabled time. Ideally, we should show alert after detector 1st enabled time, and we should not show data during time period when detector disabled. However, there is no way to keep the 1st enabled time, also no way to keep track of which time period detector is disabled. If detector is disabled/enabled multiple times, it is hard for us to ignore the disabled time periods. That's why we only show alerts since latest enabled time.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: currently with 7 days data performance seems ok. will change to use binary search or other improvement for `findTimeExistsInWindow`

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: Good catch. I guess I somehow changed code to use `getFloorPlotTime` for `existingTimes`, should use `getRoundedTimeInMin` instead.  AD result page is using rounded number for anomaly. The reason to use rounded time is that some data points does not have strict time intervals between each other, usually a small number offset in timestamp, this may cause that some data points doesn't show up in expected time period due to the offset, and then we think there is missing data in that time period.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: Using sampled feature data points in will cause that some missing data points can not be captured, because when we sample anomaly results, we collect all the data for fixed [time range](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/utils/anomalyResultUtils.ts#L314-L316), if both missing data points and existing data points are there in that range, we will not be able to catch the missing ones.

public/pages/utils/anomalyResultUtils.ts
ylwu-amzn: Will use rounded or floor time in both AD result chart and feature breakdown chart? If user deep dive into anomaly result by zooming in AD result chart and check raw data in Kibana discover, it may bring some confusion as AD result chart using rounded time. User may find raw data can't match AD result chart.

yizheliu-amazon: sorry for the confusion. Currently, rounded number is used for anomaly grade, confidence on AD result page, not for timestamp. [Source](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/utils/helpers.ts#L49) of Math.round usage. Rounded time in this PR is for alert annotation timestamp only. It won't cause any mismatch.

PR #253, opened 2020-07-16
Add progress bar for initialization

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Add progress bar for initialization

**Initializing normally:**
Screen Shot 2020-07-15 at 3 03 47 PM

**When initializing is overtime:**

Screen Shot 2020-07-16 at 6 17 26 PM

**When data is missing during initializing:**
Screen Shot 2020-07-16 at 6 53 58 PM

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -290,7 +290,10 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   ) : detector.enabled &&
                     detector.curState === DETECTOR_STATE.INIT ? (
                     <EuiHealth color={DETECTOR_STATE_COLOR.INIT}>
-                      Initializing
+                      {detector.initProgress
+                        ? //@ts-ignore
+                          `Initializing (${detector.initProgress.percentageStr} complete)`
+                        : 'Initializing'}
                     </EuiHealth>
                   ) : detector.curState === DETECTOR_STATE.INIT_FAILURE ||
                     detector.curState === DETECTOR_STATE.UNEXPECTED_FAILURE ? (
@@ -343,7 +346,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
           <EuiFlexGroup>
             <EuiFlexItem>
               <EuiTabs>
-                {tabs.map(tab => (
+                {tabs.map((tab) => (
                   <EuiTab
                     onClick={() => {
                       handleTabChange(tab.route);
@@ -388,7 +391,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   <EuiFieldText
                     fullWidth={true}
                     placeholder="delete"
-                    onChange={e => {
+                    onChange={(e) => {
                       if (e.target.value === 'delete') {
                         setDetectorDetailModel({
                           ...detectorDetailModel,
@@ -472,7 +475,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
         <Route
           exact
           path="/detectors/:detectorId/results"
-          render={props => (
+          render={(props) => (
             <AnomalyResults
               {...props}
               detectorId={detectorId}
@@ -484,7 +487,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
         <Route
           exact
           path="/detectors/:detectorId/configurations"
-          render={props => (
+          render={(props) => (
             <DetectorConfig
               {...props}
               detectorId={detectorId}
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -114,6 +114,13 @@ export function AnomalyResults(props: AnomalyResultsProps) {
   const monitors = useSelector((state: AppState) => state.alerting.monitors);
   const monitor = get(monitors, `${detectorId}.0`);
 
+  const [featureMissingSeverity, setFeatureMissingSeverity] = useState<
+    MISSING_FEATURE_DATA_SEVERITY
+  >();
+
+  const [featureNamesAtHighSev, setFeatureNamesAtHighSev] = useState(
+    [] as string[]
+  );
   const isDetectorRunning =
     detector && detector.curState === DETECTOR_STATE.RUNNING;
 
@@ -131,8 +138,16 @@ export function AnomalyResults(props: AnomalyResultsProps) {
   const isDetectorInitializing =
     detector && detector.curState === DETECTOR_STATE.INIT;
 
-  const initializationInfo = getDetectorInitializationInfo(detector);
-  const isInitOvertime = get(initializationInfo, IS_INIT_OVERTIME_FIELD, false);
+  const isDetectorMissingData = featureMissingSeverity
+    ? (isDetectorInitializing || isDetectorRunning) &&
+      featureMissingSeverity > MISSING_FEATURE_DATA_SEVERITY.GREEN
+    : undefined;
+
+  const initializationInfo = featureMissingSeverity
+    ? getDetectorInitializationInfo(detector)
+    : undefined;
+
+  const isInitOvertime = get(initializationInfo, IS_INIT_OVERTIME_FIELD);
   const initDetails = get(initializationInfo, INIT_DETAILS_FIELD, {});
   const initErrorMessage = get(initDetails, INIT_ERROR_MESSAGE_FIELD, '');
   const initActionItem = get(initDetails, INIT_ACTION_ITEM_FIELD, '');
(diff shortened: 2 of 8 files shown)


public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -343,7 +346,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
           <EuiFlexGroup>
             <EuiFlexItem>
               <EuiTabs>
-                {tabs.map(tab => (
+                {tabs.map((tab) => (
```
yizheliu-amazon, line 349: automatically changed by linter I guess. same for the other similar changes.

server/routes/ad.ts
```diff
@@ -202,6 +202,10 @@ const getDetector = async (
         ? //@ts-ignore
           { stateError: detectorState.error }
         : {}),
+      ...(detectorState !== undefined
+        ? //@ts-ignore
+          { initProgress: getDetectorInitProgress(detectorState) }
+        : {}),
```
ohltyler, line 208: Should we combine all 3 of these conditionals (setting detector `state`, `error`, `initProgress`) into one for readability?

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -290,7 +290,10 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   ) : detector.enabled &&
                     detector.curState === DETECTOR_STATE.INIT ? (
                     <EuiHealth color={DETECTOR_STATE_COLOR.INIT}>
-                      Initializing
+                      {detector.initProgress
+                        ? //@ts-ignore
+                          `Initializing (${detector.initProgress.percentageStr} complete)`
+                        : 'Initializing'}
```
ohltyler, line 296: When would the progress bar be in initializing state but not `initProgress`?

kaituo: on unpatched backend.  Only latest backend has initProgress.

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -332,6 +346,35 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
+                                color="primary"
+                                size="xs"
+                              />
+                            </EuiFlexItem>
+                          </EuiFlexGroup>
+                          <EuiSpacer size="l" />
+                        </div>
+                      ) : null}
                       <EuiButton
                         onClick={props.onSwitchToConfiguration}
```
kaituo, line 379: Why do we have a switch to configuration button when the configuration tab is only a finger click away?

ohltyler: I know this was just part of the existing approved UI, but agreed that it doesn't really seem necessary with the tab right above. Guess it was for quickly navigating to config page since a config change must have occurred for initialization process to begin again / for callout to appear again.

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -292,6 +295,17 @@ export function AnomalyResults(props: AnomalyResultsProps) {
       </p>
     ) : isInitializingNormally ? (
       <p>
+        {detector.initProgress
```
kaituo, line 298: Do we want to show the message besides isInitializingNormally?  When it is over time, I can see a need for customer to ask when it is gonna finish.

public/pages/DetectorResults/containers/AnomalyResults.tsx
yizheliu-amazon: Yeah, this already exists in UX design since long time ago. This configuration button is shown in callout because all the messages in callout mentions detector configuration, and such button can direct user quickly to detector config page .

public/pages/DetectorResults/containers/AnomalyResults.tsx
yizheliu-amazon: good point, will add it for overtime case

server/routes/ad.ts
yizheliu-amazon: good catch. will do that.

public/pages/DetectorResults/containers/AnomalyResults.tsx
yizheliu-amazon: will also add that message to missing data case as well.

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -110,6 +114,13 @@ export function AnomalyResults(props: AnomalyResultsProps) {
   const monitors = useSelector((state: AppState) => state.alerting.monitors);
   const monitor = get(monitors, `${detectorId}.0`);
 
+  const [featureMissingSeverity, setFeatureMissingSeverity] = useState<
```
yizheliu-amazon, line 117: simply move 2 lines of code from below to above.

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -273,6 +279,14 @@ export function AnomalyResults(props: AnomalyResultsProps) {
     }
   };
 
+  const getInitProgressMessage = () => {
+    return detector && isDetectorInitializing && detector.initProgress
+      ? `The detector needs to capture approximately
+                  ${detector.initProgress.neededShingles} data points for initializing. If your data stream is continuous, this process will take around
```
ylwu-amzn, line 285: How about change `${detector.initProgress.neededShingles} data points for initializing` to `${detector.initProgress.neededShingles} more data points for initializing`

ylwu-amzn: Sorry, seems I misunderstood it. Here the `${detector.initProgress.neededShingles} ` is the total data points needed or we still need such more data points?

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -273,6 +279,14 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
+                  ${detector.initProgress.estimatedMinutesLeft} minutes; if not, it may take even longer. `
```
ylwu-amzn, line 286: Just confirm, is the detector state based on how many data points in the tree? I remember our testing detector on NAB data will become running quickly before capturing 128 data points as cold start will build RCF model first. I'm asking this to make sure the detector state, neededShingles and estimatedMinutesLeft can match.

public/pages/DetectorResults/containers/AnomalyResults.tsx
kaituo: “data points” should be replaced by shingles.

kaituo: re Yaliang: we still need such more data points.

public/pages/DetectorResults/containers/AnomalyResults.tsx
kaituo: > is the detector state based on how many data points in the tree

yes

The number would be the total data points in the tree before cold start, and needed shingles after cold start.  In your NAB example, we should only see this message briefly and the progress bar should disappear as initialization is finished.

public/pages/DetectorResults/containers/AnomalyResults.tsx
yizheliu-amazon: > “data points” should be replaced by shingles.

Should we expose `shingle`, which is one internal term, to customer? I think `data points` expresses the same thing, and it is more user-friendly.

kaituo: We should expose it.  We mentioned it in our doc: https://opendistro.github.io/for-elasticsearch-docs/docs/ad/ 
 Data points and shingle are not the same thing.

yizheliu-amazon: > How about change `${detector.initProgress.neededShingles} data points for initializing` to `${detector.initProgress.neededShingles} more data points for initializing`?
> 
> Confirm with tech writer about the wording.

The message is already reviewed with tech writer. I will check with him on adding the word "more".

yizheliu-amazon: > We should expose it. We mentioned it in our doc: https://opendistro.github.io/for-elasticsearch-docs/docs/ad/
> Data points and shingle are not the same thing.

From the doc:
```
The anomaly detection model uses a shingling process, a technique that uses consecutive data points to create a sample for the model. This process needs a certain number of aggregated data points from contiguous intervals.
```
In my understanding, `shingles` are the data points we need to collect for shingling process. Is there any reason why we should not use `data points` here?

yizheliu-amazon: hmm, I see, sound like shingles are "consecutive data points", is it right?

kaituo: A shingle is a consecutive sequence of the most recent records.

yizheliu-amazon: what about using "consecutive data points", my point is that shingle is not intuitive word for user, and we only have  definition of what `shingle process` means in the [doc](https://opendistro.github.io/for-elasticsearch-docs/docs/ad/)

yizheliu-amazon: after discussing offline with Kaituo, we may remove the message regarding shingle/data point which may be confusing to user, while provide needed time and percentage only.

yizheliu-amazon: Wording has been confirmed with tech writer.

PR #266, opened 2020-07-30
Upgrade Cypress and elliptic dependencies

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR upgrades a few of the dependencies by running `yarn upgrade` commands:
1. Upgrades Cypress to 4.11.0 to remove lodash security vulnerability (see [changelog](https://docs.cypress.io/guides/references/changelog.html#4-11-0))
2. Upgrades elliptic to 6.5.3 to remove security vulnerability

Did a sanity check that plugin works fine and all UT/IT pass. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "babel-jest": "^24.8.0",
     "babel-polyfill": "^6.26.0",
     "chance": "^1.1.0",
-    "cypress": "^4.6.0",
+    "cypress": "^4.11.0",
     "eslint": "^5.6.0",
     "eslint-plugin-babel": "^5.2.0",
     "eslint-plugin-import": "^2.14.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "@elastic/charts": "^19.2.0",
     "babel-polyfill": "^6.26.0",
     "brace": "^0.11.1",
-    "elliptic": "^6.5.3",
     "eslint-plugin-cypress": "^2.11.1",
     "formik": "^1.5.8",
     "lodash": "^4.17.19",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -80,6 +80,7 @@
     "@elastic/charts": "^19.2.0",
     "babel-polyfill": "^6.26.0",
     "brace": "^0.11.1",
+    "elliptic": "^6.5.3",
```
yizheliu-amazon, line 83: Is it necessary to add direct dependency on `elliptic`? In existing repo, elliptic^6.0.0 and ^6.5.2 is resolved to 6.5.2, can you only resolve it to 6.5.3 in `yarn.lock` instead of adding direct dependency?

ohltyler: > Is it necessary to add direct dependency on `elliptic`? In existing repo, elliptic^6.0.0 and ^6.5.2 is resolved to 6.5.2, can you only resolve it to 6.5.3 in `yarn.lock` instead of adding direct dependency?

I considered that, although I just know by best practices that we are not supposed to edit `yarn.lock` directly. This was automatically populated in `package.json` when I ran `yarn upgrade elliptic@^6.5.3`. Let me know what you think.

ohltyler: Cool, so I followed the suggestion [here](https://github.com/yarnpkg/yarn/issues/4986) to manually remove the dependency in `yarn.lock` and run `yarn` right after to re-populate the dependency with an upgraded version. There is still is an issue with yarn to be able to easily upgrade indirect/sub-dependencies, thanks for calling out.

yizheliu-amazon:
Thanks for the change

PR #269, opened 2020-07-31
Remove elastic charts dependency

ohltyler:
*Issue #, if available:*

*Description of changes:*

Removes unnecessary `@elastic/charts` dependency, which is handled as a dependency in Kibana itself. Ran `yarn remove @elastic/charts`.

Confirmed the newly-fixed zoom feature works, and basic plugin functionality works. Also confirmed UT/IT pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "typescript": "3.0.3"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


yizheliu-amazon:
Cool, thanks!

PR #272, opened 2020-08-04
Add sample detectors and indices

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR adds sample detectors and indices for users to be able to quickly get started with anomaly detection, and further understand detector configurations and how detectors interact with streaming time series data.

Specific changes include:
1. Add dedicated 'Sample detectors' page in side navigation bar, which allows users to load up to 3 different sample indices and corresponding detectors.
2. Add a callout on the create detector page to guide user to the sample data (only shows up when no detectors have ever been created in the cluster before)
3. Add several internal APIs to create and bulk insert an index to ES
4. Add gzipped sample indices data on server side
5. Add corresponding UT

Have confirmed with UX on wording for all of the design.
Have confirmed all UT/IT pass

Screenshots:

Sample detectors page:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-08-12-16_01_03

Sample detectors page (one detector loaded):
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-08-12-16_01_24

Additional button on empty dashboard page:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-08-12-15_59_59

Additional button on empty detector list page:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-08-12-16_00_14

Callout on Create detector page:
Screen Shot 2020-08-12 at 4 00 51 PM


public/pages/SampleData/utils/helpers.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { DetectorListItem } from 'public/models/interfaces';
+import { SAMPLE_TYPE } from '../../../utils/constants';
+import {
+  sampleHttpResponses,
+  sampleEcommerce,
+  sampleHostHealth,
+} from '../utils/constants';
+
+export const containsDetectorsIndex = (indices: CatIndex[]) => {
+  const detectorsIndexName = '.opendistro-anomaly-detectors';
```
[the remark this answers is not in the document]
ohltyler: Good catch, this actually is already declared as a constant in public/utils/constants.ts. Will change to import that instead.

public/redux/reducers/elasticsearch.ts
```diff
@@ -204,6 +276,24 @@ export const searchES = (requestData: any): APIAction => ({
...
+    client.put(`..${AD_NODE_API.CREATE_INDEX}`, { indexConfig }),
+});
+
+export const bulk = (body: any): APIAction => ({
+  type: BULK,
+  request: (client: IHttpService) =>
+    client.post(`..${AD_NODE_API.BULK}`, { body }),
+});
+
+export const deleteIndex = (index: string): APIAction => ({
```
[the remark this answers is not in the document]
ohltyler: So currently don't support this in the UI actually. Had originally added this to support deletion, but for the current implementation we decided not to do add this on the UI. Main reason being it could get really complex and confusing for a user to know what would be deleted, and could cause errors if a user changes the name of the existing sample detector, or creates other detectors on top of the sample index, etc. Detector deletion can always be done via detector list or detector details pages, and index deletion via console.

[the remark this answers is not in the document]
ohltyler: Note that this will probably be a future improvement.

public/pages/SampleData/containers/SampleData/SampleData.tsx
```diff
@@ -0,0 +1,252 @@
...
+        console.error('Error creating sample index: ', error);
+      });
+    }
+
+    // Get the sample data from the server and bulk insert
+    if (!errorDuringAction) {
+      await dispatch(createSampleData(sampleType)).catch((error: any) => {
+        errorDuringAction = true;
+        errorMessage = error;
+        console.error('Error creating sample detector: ', error);
```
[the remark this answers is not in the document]
ohltyler: Yes, good catch, will change.

public/pages/SampleData/containers/SampleData/SampleData.tsx
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    getAllDetectors();
+    getAllIndices();
+    setLoadingState(false);
+    if (!errorDuringAction) {
+      toastNotifications.addSuccess('Successfully loaded sample detector');
+    } else {
+      toastNotifications.addDanger(
+        `Unable to load all sample data. ${errorMessage}`
```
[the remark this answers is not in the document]
ohltyler: Hard to say, since error could be from many different sources. In most cases, just trying to perform the actions again is the recommended action. Think we should change the toast to say something like `Unable to load all sample data, please try again. ${errorMessage}`

public/pages/SampleData/containers/SampleData/SampleData.tsx
```diff
@@ -0,0 +1,252 @@
...
+        })
+        .catch((error: any) => {
+          errorDuringAction = true;
+          errorMessage = error;
+          console.error('Error creating sample detector: ', error);
+        });
+    }
+
+    getAllDetectors();
+    getAllIndices();
```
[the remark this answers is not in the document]
ohltyler: Good idea, this would be optimal. Will change the params to only query for relevant detectors/indices pertaining to the sample ones.

public/pages/SampleData/containers/SampleData/SampleData.tsx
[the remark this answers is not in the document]
kaituo: Yes, this is the best we can do.  Since this may happen at any step, would it cause issues if we retry?  For example, if sample indices are already loaded, retry loading sample index would cause another error.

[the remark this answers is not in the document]
ohltyler: I handle the case where if sample indices are already loaded, we skip the step of trying to create again, and go straight to creating a detector on it (could be a common case where index and detector initially created, sample detector deleted, want to recreate the detector again).

If the error is creating the index or creating the detector, this would be something unexpected where hopefully a retry would work. At least propagating the custom error message should help the user out a bit, since the error can be from several sources.

server/routes/ad.ts
```diff
@@ -333,7 +335,7 @@ const getDetectors = async (
         query_string: {
           fields: ['name', 'description'],
           default_operator: 'AND',
-          query: `*${search.trim().split(' ').join('* *')}*`,
+          query: `*${search.trim().split('-').join('* *')}*`,
```
ohltyler, line 338: Had to change this line to delimit on `-` rather than ` ` to return results for detectors with `-`s in them. This is because the default tokenizer for ES splits up tokens on `-`.

Example: have some detector called `opendistro-sample-ecommerce-detector`, send an ES search request like `opendistro-sample*` will return empty, since it will be split into searching for `opendistro` AND `sample*`. In this case, `sample*` regex returns empty, so total result returns empty.

Since only other calls to this function send search params of `""`, changing this here doesn't break any existing functionality.

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
```diff
@@ -0,0 +1,113 @@
...
+  EuiSpacer,
+  EuiLink,
+} from '@elastic/eui';
+import { PLUGIN_NAME } from '../../../../utils/constants';
+
+interface SampleDataBoxProps {
+  title: string;
+  icon: any;
+  description: string;
+  buttonDescription: string;
```
yizheliu-amazon, line 32: maybe rename it to "LoadDataButtonDescription", "buttonDescription" sounds too generic.

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  : props.buttonDescription}
+              </EuiButton>
+            </EuiFlexItem>
+            <EuiFlexItem grow={false}>
+              {props.isDataLoaded ? (
+                <EuiLink href={`${PLUGIN_NAME}#/detectors/${props.detectorId}`}>
+                  View detector and sample data
+                </EuiLink>
+              ) : (
+                <EuiLink disabled={true}>&nbsp;</EuiLink>
```
yizheliu-amazon, line 105: Can we just return `null` instead of an disabled `EuiLink`?

server/routes/elasticsearch.ts
```diff
@@ -133,6 +136,87 @@ const getAliases = async (
...
+      index,
+      format: 'json',
+      h: 'health,index',
+    });
+    return { ok: true, response: { indices: response } };
+  } catch (err) {
+    // In case no matching indices is found it throws an error.
+    if (
+      err.statusCode === 404 &&
+      get<string>(err, 'body.error.type', '') === 'index_not_found_exception'
```
yizheliu-amazon, line 168: We may use the existing helper method: `isIndexNotFoundError`

server/routes/elasticsearch.ts
```diff
@@ -133,6 +136,87 @@ const getAliases = async (
...
+    ) {
+      return { ok: true, response: { indices: [] } };
```
yizheliu-amazon, line 170: why return OK if indexIsNotFound? Seems like creating index doesn't work es expected in this case.

server/routes/elasticsearch.ts
```diff
@@ -133,6 +136,87 @@ const getAliases = async (
...
+  console.log('index to delete: ', index);
+
+  try {
+    const response: any = await callWithRequest(req, 'indices.delete', {
+      index: index,
+    });
+    //@ts-ignore
+    return { ok: true, response: { response } };
+  } catch (err) {
+    console.log(
```
yizheliu-amazon, line 212: if indexNotFound exception is caught, should we return OK as `true`?

server/sampleData/utils/helpers.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+    });
+
+    const bulkInsert = async (docs: any[]) => {
+      try {
+        const bulkBody = prepareBody(docs, offset);
+        const resp = await callWithRequest(req, 'bulk', {
+          body: bulkBody,
+        });
+        if (resp.errors) {
```
yizheliu-amazon, line 96: log `resp.errors` and `err` before returing

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
ohltyler: sure, will change.

server/routes/ad.ts
yizheliu-amazon: I am a little confused: given delimit on `-` instead of ` ` in this line, if input search is `opendistro-sample*`, then the query becomes `*opendistro* *sample**`, will such query returns empty?

From the explanation:
```
Example: have some detector called opendistro-sample-ecommerce-detector, send an ES search request like opendistro-sample* will return empty, since it will be split into searching for opendistro AND sample*. In this case, sample* regex returns empty, so total result returns empty.
```

It looks like changing to `-` as delimiter still returns empty result. 

Btw, can you share any doc regarding `the default tokenizer for ES splits up tokens on  -`?

ohltyler: > if input search is `opendistro-sample*`, then the query becomes `*opendistro* *sample**`, will such query returns empty?

This will return results, and is the fixed behavior. Given some detector named `opendistro-sample-ecommerce-detector`, this will match the regex of both `*opendistro*` and `*sample**`.

Sorry for the confusion, but the example I wrote was in the case of splitting by ` ` rather than `-`, and was the problem I was having with returning results originally (I edited it to clarify).

As of 5.1, the default tokenizer was the standard tokenizer, which split based on common punctuation (including `-`). See [here](https://www.elastic.co/guide/en/elasticsearch/reference/5.1/analysis-standard-analyzer.html). There are several threads online of people struggling with searching with terms that include `-`, like [this one](https://discuss.elastic.co/t/how-to-search-for-terms-containing-hyphen-on--all-field/81335).

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
ohltyler: yes, will change.

server/routes/elasticsearch.ts
ohltyler: yes, thanks for this. Was copying from existing code and this was leftover. This is wrong, will fix.

server/routes/elasticsearch.ts
ohltyler: Will remove this code entirely, per below thread.

server/routes/elasticsearch.ts
ohltyler: Do you mean if an `indexNotFoundException` is thrown in the delete api call, still return ok as `true`? I feel like this should be `false` since there was an error while trying to delete, let me know what you think

server/sampleData/utils/helpers.ts
ohltyler: sure, will add

public/pages/DetectorsList/components/EmptyMessage/EmptyMessage.tsx
```diff
@@ -45,13 +51,27 @@ export const EmptyDetectorMessage = (props: EmptyDetectorProps) => (
...
         </EuiButton>
       ) : (
-        <EuiButton
-          fill
-          href={`${PLUGIN_NAME}#${APP_PATH.CREATE_DETECTOR}`}
-          data-test-subj="add_detector"
-        >
-          Create detector
-        </EuiButton>
+        <EuiFlexGroup direction="row" gutterSize="m" justifyContent="center">
```
ylwu-amzn, line 54: Looks like the same code with `public/pages/Dashboard/Components/EmptyDashboard/EmptyDashboard.tsx`. Can we remove the duplicate code ?

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'row',
+              alignContent: 'flexStart',
+            }}
+          >
+            {props.icon}
+            <EuiSpacer size="m" />
+            <h2>&nbsp;&nbsp;</h2>
```
ylwu-amzn, line 53: Why add two space here? How about adding padding?

public/components/ContentPanel/ContentPanel.tsx
```diff
@@ -45,6 +46,7 @@ const ContentPanel = (props: ContentPanelProps) => (
   <EuiPanel
     style={{ padding: '20px', ...props.panelStyles }}
     className={props.contentPanelClassName}
+    betaBadgeLabel={props.badgeLabel}
```
ylwu-amzn, line 49: Should we add some beta badge tooltip ?

public/pages/SampleData/components/SampleDataBox/SampleDataBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+              <EuiButton
+                style={{ width: '300px' }}
+                data-test-subj="loadDataButton"
+                disabled={props.isLoadingData || props.isDataLoaded}
+                isLoading={props.isLoadingData}
+                onClick={() => {
+                  props.onLoadData();
+                }}
+              >
+                {props.isLoadingData
```
ylwu-amzn, line 93: Is it possible that sample data loaded into ES index, but fail to create detector? Default maximum detector is 1000. If user have created 1000 detectors, then load sample detector will succeed, but create detector will fail. For this case, will we throw some exception ?

public/pages/createDetector/containers/CreateDetector.tsx
```diff
@@ -68,6 +73,25 @@ export function CreateDetector(props: CreateADProps) {
...
+  useEffect(() => {
+    const getInitialIndices = async () => {
+      await dispatch(getIndices(''));
+    };
+    getInitialIndices();
+  }, []);
+
+  // Check if the sample data callout should be visible based on detector index
+  useEffect(() => {
+    setSampleCalloutVisible(!containsDetectorsIndex(visibleIndices));
```
ylwu-amzn, line 93: If anomaly detector index not exists in visible indices, we will show sample detector callout on detector creation page? Why not show this callout when click "Create detector" button on dashboard/detector list page? Can you add some screenshot?

server/routes/ad.ts
ohltyler: I agree, this could cause some confusion here. Will confirm with UX

public/pages/SampleData/components/SampleDataCallout/SampleDataCallout.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import React from 'react';
+import { EuiCallOut, EuiLink } from '@elastic/eui';
+import { APP_PATH, PLUGIN_NAME } from '../../../../utils/constants';
+
+interface SampleDataCalloutProps {
+  onHide(): void;
```
yizheliu-amazon, line 21: Is `onHide()` being used?

server/routes/elasticsearch.ts
yizheliu-amazon: I personally prefer returning `true` as this can make this delete API idempotent, but this is from perspective of API, not a blocker.

ohltyler: Got it, will change to return true if an `index_not_found_exception` is returned when performing delete. Will also re-retrieve the indices if successful to properly update redux store.

public/pages/SampleData/components/SampleDataCallout/SampleDataCallout.tsx
ohltyler: No, good catch. Not necessary, will remove.

ylwu-amzn:
Got any more suggestion from UX reviewer?

ohltyler:
> Got any more suggestion from UX reviewer?

Everything has been approved by UX, tech writer, and fit n finish fixes.

ylwu-amzn:
> > Got any more suggestion from UX reviewer?
> 
> Everything has been approved by UX, tech writer, and fit n finish fixes.

Cool, good job! Will do last round check.

public/pages/SampleData/utils/constants.tsx
```diff
@@ -183,19 +183,27 @@ export const sampleEcommerce = {
...
         importance: 3,
         aggregationQuery: {
-          avg_total_revenue: { avg: { field: 'total_revenue_usd' } },
+          max_total_revenue: { max: { field: 'total_revenue_usd' } },
+        },
+      },
+      {
+        feature_name: 'min_total_revenue',
+        feature_enabled: true,
+        importance: 4,
```
ylwu-amzn, line 204: The importance was removed. Why add importance in the sample detector?

public/pages/SampleData/utils/constants.tsx
```diff
@@ -183,19 +183,27 @@ export const sampleEcommerce = {
         },
       },
       {
-        feature_name: 'sum_total_revenue',
+        feature_name: 'avg_total_revenue',
```
ylwu-amzn, line 186: So this sample detector has 4 features: `sum_failed_items`, `avg_total_revenue`, `max_total_revenue`, and `min_total_revenue`. What story we want to tell user with this sample detector?

public/pages/SampleData/utils/constants.tsx
ohltyler: Maybe I'm confused about what importance is here, I was thinking it was related to the order of features shown in the UI, but I see that it's not under uiMetadata, so maybe not. Could you explain what it means?

public/pages/SampleData/utils/constants.tsx
ohltyler: My thoughts on adding the max() and min() revenue here was to pick up on anomalous spikes related to single transactions made. The sample data can have anomalous increase spikes if an expensive transaction is made, and can have anomalous decrease spikes (such as 0) if there was failed transactions. Since there can be 10 or more transactions made per interval, the avg() isn't always best at picking this up.

Overall, the story I want to tell with this detector is find any anomalous points where extremely large purchases were made, or when there were any failed transactions.

public/pages/SampleData/utils/constants.tsx
ylwu-amzn: Original design was adding weight for feature. Feature with high importance will have more weight than others and will impact more for the anomaly result. But this design was removed later, no feature "importance" now.

ohltyler: I see, thanks for clarifying. Will remove importance for all of the features.

PR #267, opened 2020-07-31
Improve error handling when retrieving all detectors

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR improves error handling in a few different places regarding retrieving all detectors from ES. Specifically:
1. Displays a toast notification on dashboard and detector list pages if there was errors retrieving detectors
2. Fixes issue of infinite loading state on detector list page when errors retrieving detectors
3. Change to use AD redux store to identify if there was errors
4. Properly handles rejected Promises on server side to prevent the plugin from getting in a bad state (e.g., `could not retrieve property state from undefined`).

Confirmed basic functionality of plugin works, and all UT/IT pass.

Screenshots:

Screen Shot 2020-07-31 at 9 13 52 AM

Screen Shot 2020-07-31 at 9 13 29 AM

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #273, opened 2020-08-05
Handle index not found error

ohltyler:
*Issue #, if available:*

*Description of changes:*

#267 propagated a generic message to the user when there was errors retrieving detectors on the dashboard and detector list pages. One of these errors included an `index_not_found_exception`, which is most likely not an error, but is because the housekeeping AD indices are not all created until a user creates and starts a detector.

This fixes that by suppressing that error, similar to how the existing `searchDetector()` was doing it, and consequently doesn't show any errors on the dashboard and detector list pages when those indices don't exist yet.

### DIFF
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -303,10 +303,10 @@ const searchDetector = async (
       },
     };
   } catch (err) {
+    console.log('Anomaly detector - Unable to search detectors', err);
     if (isIndexNotFoundError(err)) {
       return { ok: true, response: { totalDetectors: 0, detectors: [] } };
     }
-    console.log('Anomaly detector - Unable to search detectors', err);
     return { ok: false, error: err.message };
   }
 };
@@ -512,10 +512,10 @@ const getDetectors = async (
       },
     };
   } catch (err) {
+    console.log('Anomaly detector - Unable to search detectors', err);
     if (isIndexNotFoundError(err)) {
       return { ok: true, response: { totalDetectors: 0, detectorList: [] } };
     }
-    console.log('Anomaly detector - Unable to list detectors', err);
     return { ok: false, error: err.message };
   }
 };
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -46,6 +46,7 @@ import {
   getFinalDetectorStates,
   getDetectorsWithJob,
   getDetectorInitProgress,
+  isIndexNotFoundError,
 } from './utils/adHelpers';
 import { set } from 'lodash';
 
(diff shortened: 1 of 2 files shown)


server/routes/ad.ts
```diff
@@ -302,10 +303,7 @@ const searchDetector = async (
...
     };
   } catch (err) {
-    if (
-      err.statusCode === 404 &&
-      get<string>(err, 'body.error.type', '') === 'index_not_found_exception'
-    ) {
+    if (isIndexNotFoundError(err)) {
       return { ok: true, response: { totalDetectors: 0, detectors: [] } };
     }
     console.log('Anomaly detector - Unable to search detectors', err);
```
yizheliu-amazon, line 309: nit: would prefer logging before any `return`, right after `catch`. same below.

ohltyler: > nit: would prefer logging before any `return`, right after `catch`. same below.

In this case, do you mean adding an additional log in the `if` block here? Could add

server/routes/ad.ts
```diff
@@ -302,10 +303,7 @@ const searchDetector = async (
       },
     };
   } catch (err) {
-    if (
-      err.statusCode === 404 &&
-      get<string>(err, 'body.error.type', '') === 'index_not_found_exception'
-    ) {
+    if (isIndexNotFoundError(err)) {
```
ylwu-amzn, line 306: Just confirm, this method `searchDetector ` only used on dashboard and detector list page? The index not found error is caused by detector index not created yet and this error is not visible to user?

ohltyler: > Just confirm, this method `searchDetector ` only used on dashboard and detector list page? The index not found error is caused by detector index not created yet and this error is visible to user?

so `searchDetector` is only called in the create detector page, and `getDetectors` is only called on the dashboard and detector list pages. In all of these cases, we don't want this error to be visible to the user since it only means the index hasn't been created yet, which isn't a problem.

I'm not changing any functionality in `searchDetector`, but instead just encapsulating the logic in a util function so it can be used by multiple functions

ylwu-amzn: Cool, just make sure the change will not break current design. For example, some page may need to surface the index not found error.

server/routes/ad.ts
yizheliu-amazon: no. just moving `console.log()` before `if(isIndexNotFoundError)`should be good.

ohltyler: yeah I see, will change

Issue #271, opened 2020-08-03
Add intermediate message in progress bar before cold start is complete


ohltyler:
Currently, the progress bar will show the estimated time remaining to initialize a detector. This can be confusing if this message is displayed before the cold start process is finished, where the actual initialization time may be much shorter if sufficient historical data exists.

For example, the screenshot below is a detector that will actually only take 10 minutes to initialize, but displays 1280 minutes since this is the estimated time remaining with no cold start (128 intervals):
<img width="1383" alt="Screen Shot 2020-08-03 at 10 26 31 AM" src="https://user-images.githubusercontent.com/62119629/89210577-38429900-d575-11ea-9451-3a83452c3814.png">

An [issue](https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/198) has been opened on the backend to address this by not returning the estimated time remaining until cold start has finished. When this is done, logic can be changed here to display some intermediate message in the progress bar, if the profile api call doesn't return an estimated time yet.

PR #283, opened 2020-08-23
Add intermediate callout message during cold start

ohltyler:
*Issue #, if available:* #271 

*Description of changes:*

This PR adds an intermediate callout to the detector anomaly results page when the detector is still in the middle of the cold start process. This fixes the bug of displaying a (possibly) very incorrect time estimation regarding detector initialization, where the time may be much shorter if sufficient historical data exists.

**NOTES**:
- Backend has been correspondingly changed, and will return undefined for the estimated time if the cold start process has not finished.
- Will confirm with UX on layout/wording.
- Confirmed all UT pass.

Specific changes:
1. Add intermediate callout above anomaly results detail page when cold start process is happening. The estimated time for this to finish is one detector interval, and is the best estimate we can provide.
2. Removing the percentage string in the detector state at the top of the detector details pages if cold start process is happening.

Before:
Screen Shot 2020-08-23 at 11 18 05 AM

After:
Screen Shot 2020-08-26 at 4 13 26 PM

### DIFF
--- a/public/pages/DetectorDetail/containers/DetectorDetail.tsx
+++ b/public/pages/DetectorDetail/containers/DetectorDetail.tsx
@@ -290,7 +290,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   ) : detector.enabled &&
                     detector.curState === DETECTOR_STATE.INIT ? (
                     <EuiHealth color={DETECTOR_STATE_COLOR.INIT}>
-                      {detector.initProgress
+                      {detector.initProgress?.estimatedMinutesLeft
                         ? //@ts-ignore
                           `Initializing (${detector.initProgress.percentageStr} complete)`
                         : 'Initializing'}
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -269,7 +269,7 @@ export function AnomalyResults(props: AnomalyResultsProps) {
               <p>
                 Attempting to initialize the detector with historical data. This
                 will take approximately{' '}
-                {detector.detectionInterval.period.interval} total minutes.
+                {detector.detectionInterval.period.interval} minutes.
               </p>
             </EuiText>
           </EuiFlexGroup>
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -254,7 +254,27 @@ export function AnomalyResults(props: AnomalyResultsProps) {
       );
     }
     if (isPerformingColdStart) {
-      return `Attempting to initialize the detector with historical data. This will take approximately ${detector.detectionInterval.period.interval} total minutes.`;
+      return (
+        <div>
+          <EuiFlexGroup direction="row" gutterSize="s">
+            <EuiLoadingSpinner
+              size="l"
+              style={{
+                marginLeft: '4px',
+                marginRight: '8px',
+                marginBottom: '8px',
+              }}
+            />
+            <EuiText>
+              <p>
+                Attempting to initialize the detector with historical data. This
+                will take approximately{' '}
+                {detector.detectionInterval.period.interval} total minutes.
+              </p>
+            </EuiText>
+          </EuiFlexGroup>
+        </div>
+      );
     }
     if (isInitializingNormally) {
       return 'The detector is being initialized based on the latest configuration changes.';
@@ -355,16 +375,18 @@ export function AnomalyResults(props: AnomalyResultsProps) {
                     <EuiCallOut
                       title={getCalloutTitle()}
                       color={getCalloutColor()}
-                      iconType={isInitializingNormally ? 'iInCircle' : 'alert'}
+                      iconType={
+                        isPerformingColdStart
+                          ? ''
+                          : isInitializingNormally
+                          ? 'iInCircle'
+                          : 'alert'
+                      }
                       style={{ marginBottom: '20px' }}
                     >
                       {getCalloutContent()}
-                      {isPerformingColdStart ? (
-                        <div>
-                          <EuiLoadingSpinner size="l" />
-                          <EuiSpacer size="s" />
-                        </div>
-                      ) : isDetectorInitializing && detector.initProgress ? (
+                      {isPerformingColdStart ? null : isDetectorInitializing &&
+                        detector.initProgress ? (
                         <div>
                           <EuiFlexGroup alignItems="center">
                             <EuiFlexItem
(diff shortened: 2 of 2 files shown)


ohltyler:
Will confirm with UX on layout/wording.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -290,7 +290,7 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
                   ) : detector.enabled &&
                     detector.curState === DETECTOR_STATE.INIT ? (
                     <EuiHealth color={DETECTOR_STATE_COLOR.INIT}>
-                      {detector.initProgress
+                      {detector.initProgress?.estimatedMinutesLeft
```
ylwu-amzn, line 293: `estimatedMinutesLeft` means how many minutes user needs to wait for initialization to complete? For example, if user need to wait for another 5 minutes to pass initialization, `estimatedMinutesLeft` will be 5?

ohltyler: `estimatedMinutesLeft` is calculated in the backend, which is num shingles needed * detector interval. Returns undefined until cold start is passed.

More details [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/210)

ohltyler:
@yizheliu-amazon will confirm with tech writer on wording. Will wait to merge until then.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ylwu-amzn: Never mind, find answer in `AnomalyResults.tsx` file. We will split cases before/after cold start and show different messages.

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -243,6 +253,29 @@ export function AnomalyResults(props: AnomalyResultsProps) {
...
+                marginLeft: '4px',
+                marginRight: '8px',
+                marginBottom: '8px',
+              }}
+            />
+            <EuiText>
+              <p>
+                Attempting to initialize the detector with historical data. This
+                will take approximately{' '}
+                {detector.detectionInterval.period.interval} total minutes.
```
ylwu-amzn, line 272: This is true if source index has enough data. How about the case that source index has no enough data ?

ohltyler: Wouldn't the cold start process time be pretty independent of the amount of data in the source index? It either finds data for the model, or doesn't, but will still finish quick (I've never been able to make the cold start process last longer than a detector interval).

If there's not enough data, the cold start process will still finish, but then we can use the progress bar for displaying the estimated time left to finish initializing.

This is just an estimate, I figured it is better to provide some estimate rather than none at all.

ylwu-amzn: The cold start process will finish after 1 interval. But it may not succeed if there is no historical data. Then AD will rerun cold start in next interval until find some historical data and cold start run successfully.

I agree with you, as this is just an estimate. I see you will confirm with tech writer about the wording. Just make sure tech writer understand this case.

ohltyler: So not sure I understand, are you saying that cold start will be re-ran every interval if the detector is still initializing then?

ylwu-amzn: If no historical data, cold start process ends but RCF model is not available. For this case, will rerun cold start in next interval and detector is still initializing.

ohltyler: Oh, was not aware this was how this worked. I thought cold start was triggered once when the detector was first created, and not every interval where there wasn't sufficient historical data. This kind of defeats the entire purpose of adding this change then, since it will always show this callout until there is enough data, then go straight to running state.

Seems like the progress bar will never be shown. Am I understanding this correctly?

ylwu-amzn: My understanding 
1. If there is enough historical data for cold start and shingle find enough consecutive historical data points, then the progress bar show as 100% after cold start and shingle
2. If there is enough historical data for cold start and shingle doesn't find enough consecutive historical data points, then the progress bar show as X% after cold start, X=(shinlge size) / DETECTOR_SHINLGE_SIZE

@kaituo  can you confirm the cold start and shingle logic?

yizheliu-amazon: My thoughts on this:

In case of `estimatedTime` unavailable, we may not need to add "This will take approximately X total minutes." Just say "Data is being collected to initialize model." in the intermediate callout. If `estimatedTime` is available, then show existing callout with progress bar.

kaituo: @ylwu-amzn 
X = completed shingles / required shingles

kaituo: > Oh, was not aware this was how this worked. I thought cold start was triggered once when the detector was first created, and not every interval where there wasn't sufficient historical data. This kind of defeats the entire purpose of adding this change then, since it will always show this callout until there is enough data, then go straight to running state.
> 
> Seems like the progress bar will never be shown. Am I understanding this correctly?
Your callout is necessary.  Cold start can finish without fully initializing the model.  What Yaliang said is the case there is no complete shingle in the history at all.  For example, if there is only one full shingle, cold start will finish with that.  We still need 127 shingles.  So the progress percentage is 1/128.

kaituo: > My thoughts on this:
> 
> In case of `estimatedTime` unavailable, we may not need to add "This will take approximately X total minutes." Just say "Data is being collected to initialize model." in the intermediate callout. If `estimatedTime` is available, then show existing callout with progress bar.

This can solve Yaliang's concern: cold start may not succeed.   Tyler said "approximately".  So it should cover cold start may not succeed case.  I am fine with both solutions: leave it as it is, or remove the cold start estimated finish time.

ohltyler: Got it. That makes sense. I am fine for either soln as well. Could also change to give an estimation range, such as [interval - 6*interval], although this may be exposing too much of an internal implementation detail about the initialization process, and would have to be maintained.

I would like to assume in the majority of cases, the detector is detecting over an index with at least some data, in which case I think having an estimate could be much more helpful here.

ylwu-amzn: > I would like to assume in the majority of cases, the detector is detecting over an index with at least some data, in which case I think having an estimate could be much more helpful here.

Agree, should be an edge case that no historical data to do cold start.

ohltyler:
Confirmed with tech writer to keep estimation, but remove the word `total`. Changes have been pushed and screenshot updated in description.

Issue #282, opened 2020-08-23
Number field not show up when use average on edit feature page


ylwu-amzn:
Number fields which defined in [`fielddata`](https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html#before-enabling-fielddata) will be ignored in current implementation. 

We need to parse types in `fielddata` and show number fields if user choose average/sum/max/min.

PR #284, opened 2020-08-23
parse types in fielddata

ylwu-amzn:
*Issue #282 *

## Description of changes:

We don't parse types in `fielddata` currently, so if user choose averge/sum/max/min on edit feature page, some number fields will not show up on edit feature page.

Create test index
```
PUT test_index_num_filed
{
  "mappings": {
    "properties": {
      "test": {
        "properties": {
          "a": {
            "properties": {
              "b": {
                "type": "keyword",
                "eager_global_ordinals": true,
                "fields": {
                  "raw": {
                    "type": "integer"
                  }
                }
              }
            }
          }
        }
      },
      "timestamp": {
        "type": "date",
        "format": "strict_date_time||epoch_millis"
      },
      "value": {
        "type": "float"
      }
    }
  }
}

PUT test_index_num_filed/_doc/1
{
  "test": {
    "a": {
      "b": 1
    }
  },
  "timestamp": 1587344356869,
  "value": 1
}
```
Create detector with test index `test_index_num_filed`. The edit feature page can show more number fields after change.
### Before

Screen Shot 2020-08-23 at 11 20 36 AM

### After
Screen Shot 2020-08-23 at 11 26 22 AM

### DIFF
--- a/public/redux/reducers/mapper.ts
+++ b/public/redux/reducers/mapper.ts
@@ -63,7 +63,6 @@ export function getTypeFromMappings(
         resolvePath(path, field)
       );
     });
-    return currentDataTypes;
   }
 
   return currentDataTypes;
--- a/public/redux/reducers/mapper.ts
+++ b/public/redux/reducers/mapper.ts
@@ -54,6 +54,18 @@ export function getTypeFromMappings(
   } else {
     currentDataTypes[type] = [path];
   }
+
+  if (mappings.fields) {
+    Object.entries(mappings.fields).forEach(([field, value]) => {
+      currentDataTypes = getTypeFromMappings(
+        value as Mappings,
+        currentDataTypes,
+        resolvePath(path, field)
+      );
+    });
+    return currentDataTypes;
+  }
+
   return currentDataTypes;
 }
 
--- a/public/redux/reducers/__tests__/mapper.test.ts
+++ b/public/redux/reducers/__tests__/mapper.test.ts
@@ -1,0 +1,49 @@
+import { getPathsPerDataType } from '../mapper';
+
+describe('mapper', () => {
+  describe('getPathsPerDataType', () => {
+    test('has fields', async () => {
+      const mappings = {
+        test_index: {
+          mappings: {
+            properties: {
+              test: {
+                properties: {
+                  a: {
+                    properties: {
+                      b: {
+                        type: 'keyword',
+                        eager_global_ordinals: true,
+                        fields: {
+                          raw: {
+                            type: 'integer',
+                          },
+                        },
+                      },
+                    },
+                  },
+                },
+              },
+              timestamp: {
+                type: 'date',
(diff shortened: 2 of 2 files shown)


ylwu-amzn:
UT passed locally
```
Test Suites: 1 skipped, 47 passed, 47 of 48 total
Tests:       4 skipped, 220 passed, 224 total
Snapshots:   35 passed, 35 total
Time:        254.927s
Ran all test suites.
✨  Done in 257.77s.
```

IT passed localled
```
       Spec                                              Tests  Passing  Failing  Pending  Skipped
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  ad/dashboard/ad_dashboard.spec.ts        01:12        6        6        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  ad/detectorList/detector_list.spec.      02:06       10       10        -        -        - │
  │    ts                                                                                          │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  ad/workflow/create_detector.spec.ts      00:34        1        1        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        03:53       17       17        -        -        -
```

public/redux/reducers/mapper.ts
```diff
@@ -54,6 +54,18 @@ export function getTypeFromMappings(
...
+    Object.entries(mappings.fields).forEach(([field, value]) => {
+      currentDataTypes = getTypeFromMappings(
+        value as Mappings,
+        currentDataTypes,
+        resolvePath(path, field)
+      );
+    });
+    return currentDataTypes;
+  }
+
```
ohltyler, line 68: So if I understand correctly, this is just recursively retrieving types from any nested mappings?

ylwu-amzn: Not any nested mappings, this block will parse only `fields`. In my testing index, it will parse this part
```
"fields": {
    "raw": {
         "type": "integer"
    }
}
```

ohltyler: Got it, ok, makes sense.

public/redux/reducers/mapper.ts
```diff
@@ -54,6 +54,18 @@ export function getTypeFromMappings(
...
+  if (mappings.fields) {
+    Object.entries(mappings.fields).forEach(([field, value]) => {
+      currentDataTypes = getTypeFromMappings(
+        value as Mappings,
+        currentDataTypes,
+        resolvePath(path, field)
+      );
+    });
+    return currentDataTypes;
```
yizheliu-amazon, line 66: minor: this `return` might not be necessary as next line is the same.

ylwu-amzn: Good catch, thanks.

PR #287, opened 2020-08-27
Adds window size as advanced setting in model configuration.

LiuJoyceC:
*Description of changes:*

This change allows window size (shingle size) to be configured on a detector. The modifications follow the proposed UX design. The Features page is now a more general model configuration page, which includes an Advanced settings section where the window size can be configured.

This is an initial commit intended for review of the UI (visually and functionally). A follow-up commit will refactor (since the scope of the original Features page has now expanded to have multiple sections, multiple components/files will need to be renamed as well as refactored to be more modular) and add new tests. Separating the changes into multiple commits and stages of revision has the benefits:
1) Review of the UI can begin early and not be delayed by refactoring/testing.
2) Since frontend changes generally have more subjectivity and can have significantly more parts moved around during review (unlike a backend API change which has a more clearly defined change in interface/behavior), it makes sense for the UI to get some initial approval/finalization before doing refactoring and testing (which, once the UI is finalized, should not affect the UI visually and functionally).
3) Since files need to be refactored and renamed, putting all of these changes in one commit will make it difficult to actually see a true diff of what changed vs. what just got moved to a different file. Ideally, refactors that involve renaming and moving of code should be in separate commits from commits where code content actually changed.

### DIFF
--- a/public/pages/DetectorConfig/containers/Features.tsx
+++ b/public/pages/DetectorConfig/containers/Features.tsx
@@ -20,6 +20,7 @@ import {
   EuiIcon,
   EuiButton,
   EuiEmptyPrompt,
+  EuiSpacer,
 } from '@elastic/eui';
 import {
   Detector,
@@ -27,7 +28,7 @@ import {
   FeatureAttributes,
 } from '../../../models/interfaces';
 import { get, sortBy } from 'lodash';
-import { PLUGIN_NAME } from '../../../utils/constants';
+import { PLUGIN_NAME, SHINGLE_SIZE } from '../../../utils/constants';
 import ContentPanel from '../../../components/ContentPanel/ContentPanel';
 import { CodeModal } from '../components/CodeModal/CodeModal';
 import { getTitleWithCount } from '../../../utils/utils';
@@ -93,6 +94,7 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
 
   public render() {
     const featureAttributes = get(this.props.detector, 'featureAttributes', []);
+    const shingleSize = get(this.props.detector, 'shingleSize', SHINGLE_SIZE);
 
     const sorting = {
       sort: {
@@ -170,7 +172,7 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
       },
       {
         field: 'state',
-        name: 'State',
+        name: 'Feature state',
       },
     ];
 
@@ -182,17 +184,21 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
 
     const featureNum = Object.keys(featureAttributes).length;
 
+    const setParamsText = `Set the index fields that you want to find anomalies for by defining
+                           the model features. You can also set other model parameters such as
+                           window size.`
+
+    const previewText = `After you set the model features and other optional parameters, you can
+                         preview your anomalies from a sample feature output.`
+
     return (
       <ContentPanel
-        title={getTitleWithCount('Features', featureNum)}
+        title="Model configuration"
         titleSize="s"
         subTitle={
           <EuiText className="anomaly-distribution-subtitle">
             <p>
-              Specify index fields that you want to find anomalies for by
-              defining features. A detector can discover anomalies for up to 5
-              features. Once you define the features, you can preview your
-              anomalies from a sample feature output.{' '}
+              {`${setParamsText} ${previewText} `}
               <EuiLink
                 href="https://opendistro.github.io/for-elasticsearch-docs/docs/ad/"
                 target="_blank"
@@ -211,14 +217,15 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
           <EuiEmptyPrompt
             title={
               <span className="emptyFeatureTitle">
-                Features are required to run a detector
+                Model parameters are required to run a detector
               </span>
             }
             body={
               <EuiText className="emptyFeatureBody">
-                Specify index fields that you want to find anomalies for by
-                defining features. Once you define the features, you can preview
-                your anomalies from a sample feature output.
+                {setParamsText}
+                <br/>
+                <br/>
+                {previewText}
               </EuiText>
             }
             actions={[
@@ -227,18 +234,37 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
                 href={`${PLUGIN_NAME}#/detectors/${this.props.detectorId}/features`}
                 fill
               >
-                Add feature
+                Configure model
               </EuiButton>,
             ]}
           />
         ) : (
-          <EuiBasicTable
-            items={sortedItems}
-            columns={columns}
-            cellProps={getCellProps}
-            sorting={sorting}
-            onChange={this.handleTableChange}
-          />
+          <div>
+            <ContentPanel
+              title={getTitleWithCount('Features', featureNum)}
+              titleSize="s"
+            >
+              <EuiBasicTable
+                items={sortedItems}
+                columns={columns}
+                cellProps={getCellProps}
+                sorting={sorting}
+                onChange={this.handleTableChange}
+              />
+            </ContentPanel>
+            <EuiSpacer size="m"/>
+            <ContentPanel
+              title="Additional settings"
+              titleSize="s"
+            >
+              <EuiBasicTable
+                className="header-single-value-euiBasicTable"
+                items={[{ windowSize: shingleSize }]}
+                columns={[{ field: 'windowSize', name: 'Window size'}]}
+                cellProps={getCellProps}
+              />
+            </ContentPanel>
+          </div>
         )}
       </ContentPanel>
     );
--- a/public/pages/EditFeatures/containers/EditFeatures.tsx
+++ b/public/pages/EditFeatures/containers/EditFeatures.tsx
@@ -29,9 +29,11 @@ import {
   EuiIcon,
   EuiCallOut,
   EuiSpacer,
+  EuiFieldNumber,
+  EuiFormRow,
 } from '@elastic/eui';
-import { FieldArray, FieldArrayRenderProps, Form, Formik } from 'formik';
-import { get, isEmpty, forOwn } from 'lodash';
+import { FieldArray, FieldArrayRenderProps, Form, Formik, Field, FieldProps } from 'formik';
+import { get, isEmpty } from 'lodash';
 import React, { Fragment, useState, useEffect, useCallback } from 'react';
 import { useDispatch } from 'react-redux';
 import { RouteComponentProps } from 'react-router-dom';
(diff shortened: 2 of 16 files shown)


ylwu-amzn:
Hi, Joyce, thanks for the change. Can you add some screenshot here? So others can easily see the new UI.

LiuJoyceC:
@ylwu-amzn Sure, I will add screenshots. Thanks for the suggestion.

LiuJoyceC:
Here are screenshots of the affected pages. Note that the Model configuration page looks small because I had to zoom out in the browser in order for the whole page to be seen.

Screen Shot 2020-08-27 at 11 28 02 AM
Screen Shot 2020-08-27 at 11 29 13 AM
Screen Shot 2020-08-27 at 11 29 49 AM
Screen Shot 2020-08-27 at 11 30 27 AM

public/pages/DetectorConfig/containers/Features.tsx
```diff
@@ -182,17 +187,21 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
 
     const featureNum = Object.keys(featureAttributes).length;
 
+    const setParamsText = `Set the index fields that you want to find anomalies for by defining
+                           the model features. You can also set other model parameters such as
+                           window size.`
+
+    const previewText = `After you set the model features and other optional parameters, you can
+                         preview your anomalies from a sample feature output.`
```
yizheliu-amazon, line 195: Are these wordings reviewed by tech writer?

Also, since those are constants, you may name it like SET_PARAMS_TEXT = '....'

public/pages/DetectorConfig/containers/Features.tsx
```diff
@@ -93,6 +97,7 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
 
   public render() {
     const featureAttributes = get(this.props.detector, 'featureAttributes', []);
+    const shingleSize = get(this.props.detector, 'shingleSize', SHINGLE_SIZE);
```
ylwu-amzn, line 100: Is `shingleSize` optional for backend Java Detector class? If yes, suggest to set `shingleSize=8` for java detector instance. 
So if we change the default SHINGLE_SIZE in future, we don't need to worry about the backward compatibility about old detectors. For example, user create detector with default SHINGLE_SIZE 8 now, later we change the default SHINGLE_SIZE as 4. If we don't set this detector's shingle size as 8 in data model, we may reset it as 4, and model will fail as checkpoint can't match current detector shingle size configuration.

public/pages/DetectorConfig/containers/Features.tsx
```diff
@@ -227,18 +237,34 @@ export class Features extends Component<FeaturesProps, FeaturesState> {
...
+                cellProps={getCellProps}
+                sorting={sorting}
+                onChange={this.handleTableChange}
+              />
+            </EuiFlexItem>
+            <EuiSpacer size="m"/>
+            <EuiFlexItem>
+              <EuiTitle size="s"><h3>Advanced settings</h3></EuiTitle>
+              <EuiHorizontalRule margin="xs"/>
+              <EuiBasicTable
```
ylwu-amzn, line 261: We use `EuiBasicTable` as we have multiple features, so we can put features in separate row.  Should we use the same UX design for `Advanced setting` ? How about use the similar design of `Detector configuration` ? Can you confirm with UX designer?

public/pages/EditFeatures/containers/EditFeatures.tsx
```diff
@@ -278,12 +343,20 @@ export function EditFeatures(props: EditFeaturesProps) {
                   </ContentPanel>
                 </EuiPageBody>
               </EuiPage>
+              <EuiPage>
+                <EuiPageBody>
+                  <ContentPanel title="Advanced Settings" subTitle={renderAdvancedSettingsToggle()}>
```
ylwu-amzn, line 348: Should we show some description below the `Advanced Settings`?

public/pages/EditFeatures/containers/EditFeatures.tsx
```diff
@@ -243,11 +247,72 @@ export function EditFeatures(props: EditFeaturesProps) {
...
+              <EuiFieldNumber
+                id="shingleSize"
+                placeholder="Window size"
+                data-test-subj="shingleSize"
+                {...field}
+              />
+            </EuiFlexItem>
+            <EuiFlexItem>
+              <EuiText>
+                <p className="minutes">points</p>
```
ylwu-amzn, line 299: Feel `points` not so easy for new user to understand. Consult tech writer about this. If we use`points`, suggest to add some description about what "points" means. One point means features' aggregation results for one detection interval.

public/pages/EditFeatures/containers/EditFeatures.tsx
```diff
@@ -243,11 +247,72 @@ export function EditFeatures(props: EditFeaturesProps) {
     }
   };
 
+  const renderAdvancedSettingsToggle = () => (
+    <EuiText
+      className="content-panel-subTitle"
+      onClick={()=>{setShowAdvancedSettings(!showAdvancedSettings);}}
+    >
+      <EuiLink>
+        {showAdvancedSettings ? 'Hide' : 'Show'}
```
ylwu-amzn, line 256: For a new detector, it's ok to hide the window size setting to keep the UI simple and don't bring too much learning burden about the advanced setting. But if user choose to change the default shingle size and update the detector, why we still hide this window size ? For example, user create a detector, then go to this model configuration page, add one feature and set window size as 4, should we still hide the window size for this detector when user edit it later?

public/pages/DetectorConfig/containers/Features.tsx
LiuJoyceC: The wording is based exactly on the UX mockups. I can ask the UX designer if the mockups were reviewed by a tech writer.

Sure, I will change the variable names for these constants.

public/pages/DetectorConfig/containers/Features.tsx
LiuJoyceC: I'll reach out to the UX designer to clarify. In the mockups, the `Advanced setting` section of this panel is stylistically identical to the features table in this panel.

public/pages/EditFeatures/containers/EditFeatures.tsx
LiuJoyceC: Do you mean in addition to the description below in the window size configuration. Currently, the window size config is the only thing in the Advanced settings section, so I'm not sure it makes sense to add a description both under the Advanced settings and under the window size config.

ylwu-amzn: Make sense, let's keep it as is.

public/pages/DetectorConfig/containers/Features.tsx
LiuJoyceC: The UX designer confirmed that the mockups have been reviewed by a tech writer.

public/pages/EditFeatures/containers/EditFeatures.tsx
LiuJoyceC: The UX designer confirmed that this was reviewed by a tech writer. The description below the config box provides more context to the user.

I can relay your feedback to the UX designer if you feel that it is still not clear with the description.

public/pages/DetectorConfig/containers/Features.tsx
LiuJoyceC: Could you clarify what change you're suggesting in the frontend code? Are you suggesting the SHINGLE_SIZE constant be removed from the frontend?

The `shingleSize` parameter is optional in the API, and the backend already defaults the shingle size to 8 if the parameter is not provided in the request. However, this change will only be applied to detectors which get created or updated after this backend change was made.

The reason the default shingle size is also specified in the frontend is precisely for backwards-compatibility. Detectors that were created prior to the recent backend change and have not been updated since then will not have a "shingleSize" field in the stored detector configuration, and it would not be a good user experience to see a "null" or blank in the Window Size field when they view their model configurations in the UI. Even if there is no "shingleSize" field stored on the detector configuration, the user should be informed by the UI that a shingle size of 8 is being used for that detector. In addition, the shingle size needs to be known by the frontend to calculate detector initialization time (see `isDetectorInitOverTime`). (The SHINGLE_SIZE constant was actually already existing in the code for this reason. This constant did not get added in this change.)

If they save any changes on the model configuration page for the existing detector, even if they didn't change the window size, going forward the "shingleSize" field will get stored on the detector configuration for the future, but until all existing detectors have been updated to have that field, the default value is needed to ensure the model configuration display the correct value and calculates the correct detector initialization time.

ylwu-amzn: As I said "suggest to set shingleSize=8 for java detector instance." now you have already set it as 8, "the backend already defaults the shingle size to 8 if the parameter is not provided in the request". So that will be ok.

public/pages/EditFeatures/containers/EditFeatures.tsx
ylwu-amzn: > The UX designer confirmed that this was reviewed by a tech writer. The description below the config box provides more context to the user.

Cool, thanks. I'm ok if tech writer confirms.

public/pages/DetectorConfig/containers/Features.tsx
LiuJoyceC: I synced up with the UX designer, and we will change the Advanced settings to be a content panel nested inside the Model configuration panel rather than a table.

public/pages/EditFeatures/containers/EditFeatures.tsx
LiuJoyceC: I synced up with the tech writer, and we decided to change "points" (and "data points" in the description) to "intervals", since this more accurately describes what the window size is referring to.

public/pages/EditFeatures/containers/EditFeatures.tsx
LiuJoyceC: I discussed this with the UX designer and we decided to keep this behavior. There isn't an easy way to tell if a user has explicitly set the window size on a detector in the past without storing this information in the ES detectors index, and it doesn't make much sense to modify the backend to store information that is purely a UX concern. Furthermore, the UX designer said the current behavior is consistent with advanced configuration panels in the AWS console.

public/pages/EditFeatures/containers/EditFeatures.tsx
ylwu-amzn: I think "intervals" is better. Thanks.

public/pages/DetectorConfig/containers/Features.tsx
ylwu-amzn: Cool, content panel should be good. Be sure to put both features and advanced setting to nested content panels.

public/pages/EditFeatures/containers/EditFeatures.tsx
ylwu-amzn: Thanks for confirming with UX designers. Should be easy if we show the advanced setting if its value is different with default value. 
Not a big impact to user experience for showing/hiding advanced setting if detector not using default setting. Respect UX designer's choice. Let's keep it as is.

LiuJoyceC:
Screen Shot 2020-09-01 at 8 49 31 PM
Screen Shot 2020-09-01 at 8 50 02 PM

losifer:
@LiuJoyceC could we look into moving the Window size description help text between the title label and the dropdown form?  Text below the input field should be reserved for constraints and restrictions within the input. I see though that within the Features this is also placed below the fields, I'll see what I can do to bring that updated.

ylwu-amzn:
Suggest to keep every commit, don't squash, so reviewer can review the new commit easily, otherwise, need to go through all files.

ylwu-amzn (approved):
LGTM. Thanks for the change.

Issue #299, opened 2020-09-03
Initialization callout is sometimes out of date when first loading


ohltyler:
Currently, when newly creating a detector and visiting the anomaly results page for a detector, the displayed callout is sometimes out of date and showing no callout related to any of the init progress, if applicable. For example, when creating a sample detector and clicking on the link to go to the detector details, the page shows this:

<img width="1416" alt="Screen Shot 2020-09-03 at 9 14 22 AM" src="https://user-images.githubusercontent.com/62119629/92140817-732f2b00-edc6-11ea-9aef-1fd011896b10.png">

Investigate why this is happening.

PR #300, opened 2020-09-03
Fix initialization callouts to properly show when first loading anomaly results page

ohltyler:
*Issue #, if available:* #299 

*Description of changes:*

This PR fixes the issue of the anomaly results page not properly showing the callouts relating to init progress (if applicable) when first loading the page after detector creation. Specifically, this fix will re-get the detector details when first rendering the page, to update the redux store with the detector details if they haven't been updated already.

The issue (before): user creates detector -> redux store gets updated with detector info returned by create detector api (which doesn't include any info related to init progress) -> user clicks on anomaly results page -> page uses existing detector info -> assumes no init progress -> renders old callout. Note that this issue only comes up if the store hasn't already been updated by rendering some other page, like detector list or dashboard for example, where getDetectors() is called, which would update the redux store with up-to-date detector details.

Confirmed all UT/IT pass.

Will cherry-pick to opendistro-1.10 branch and push new tag.

Screenshots:

Before:
Screen Shot 2020-09-03 at 9 14 22 AM

After:
Screen Shot 2020-09-03 at 9 16 44 AM

### DIFF
--- a/public/pages/DetectorResults/containers/AnomalyResults.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyResults.tsx
@@ -88,10 +88,11 @@ export function AnomalyResults(props: AnomalyResultsProps) {
       BREADCRUMBS.DETECTORS,
       { text: detector ? detector.name : '' },
     ]);
+    dispatch(getDetector(detectorId));
   }, []);
 
   const fetchDetector = async () => {
-    await dispatch(getDetector(detectorId));
+    dispatch(getDetector(detectorId));
   };
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -88,10 +88,11 @@ export function AnomalyResults(props: AnomalyResultsProps) {
       BREADCRUMBS.DETECTORS,
       { text: detector ? detector.name : '' },
     ]);
+    dispatch(getDetector(detectorId));
   }, []);
 
   const fetchDetector = async () => {
-    await dispatch(getDetector(detectorId));
+    dispatch(getDetector(detectorId));
```
yizheliu-amazon, line 95: wondering why removing `await` here?

yizheliu-amazon (approved):
thanks for the fix

public/pages/DetectorResults/containers/AnomalyResults.tsx
ohltyler: Unnecessary, since dispatch() is just used to trigger a separate async action to update the redux store, and the fn itself is already async. Also, nothing is done after the action is done within this function (like catch errors, do something with the response, etc). VS code also complains about it being unnecessary / has no effect.

Once the action is done, the page will render with the updated `detector` since the redux store will be updated.

yizheliu-amazon:
```
VS code also complains about it being unnecessary / has no effect.
```
Hmm, why I didn't see that in my VS code. is it because of any good plugin?

ohltyler: Not sure, don't have extensions specific to javascript/typescript/react. Just looks like this. Have also noticed it in many other places, just fixed here though.

Screen Shot 2020-09-03 at 11 35 02 AM

ylwu-amzn (approved):

PR #305, opened 2020-09-04
Fix 2 issues on Dashboard

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Issue 1: Show live chart even when live anomaly data is empty

**Before**
Screen Shot 2020-09-04 at 9 48 46 AM

**After** 
Screen Shot 2020-09-04 at 9 48 16 AM

Issue 2: Fix full screen issue for Dashboard

**Before**
Screen Shot 2020-09-04 at 11 23 09 AM

**After**
Screen Shot 2020-09-04 at 11 22 24 AM

### DIFF
--- a/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
@@ -57,7 +57,7 @@ import {
   getLatestAnomalyResultsForDetectorsByTimeRange,
   getLatestAnomalyResultsByTimeRange,
 } from '../utils/utils';
-import { MAX_ANOMALIES } from '../../../utils/constants';
+import { MAX_ANOMALIES, SPACE_STR } from '../../../utils/constants';
 
 export interface AnomaliesLiveChartProps {
   selectedDetectors: DetectorListItem[];
@@ -134,7 +134,7 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
 
     setLatestLiveAnomalousDetectorsCount(
       new Set(
-        latestLiveAnomalyResult.map(anomalyData =>
+        latestLiveAnomalyResult.map((anomalyData) =>
           get(anomalyData, AD_DOC_FIELDS.DETECTOR_ID, '')
         )
       ).size
@@ -165,7 +165,7 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
     liveTimeRange.endDateTime.valueOf(),
   ]);
 
-  const visualizedAnomalies = liveAnomalyData.flatMap(anomalyResult =>
+  const visualizedAnomalies = liveAnomalyData.flatMap((anomalyResult) =>
     visualizeAnomalyResultForXYChart(anomalyResult)
   );
   const prepareVisualizedAnomalies = (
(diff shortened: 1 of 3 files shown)


public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
```diff
@@ -187,12 +187,13 @@ export const AnomaliesLiveChart = (props: AnomaliesLiveChartProps) => {
         continue;
       }
       result.push({
-        [AD_DOC_FIELDS.DETECTOR_NAME]: !isEmpty(liveAnomalyData) ? '' : null,
+        [AD_DOC_FIELDS.DETECTOR_NAME]: !isEmpty(liveAnomalyData)
+          ? ''
+          : SPACE_STR,
```
yizheliu-amazon, line 192: This is the only change for fix on the live chart issue

public/pages/Dashboard/Container/DashboardOverview.tsx
```diff
@@ -203,75 +203,77 @@ export function DashboardOverview() {
...
-              />
-            </EuiFlexItem>
-            <EuiFlexItem grow={3}>
-              <AnomalousDetectorsList selectedDetectors={currentDetectors} />
-            </EuiFlexItem>
-          </EuiFlexGroup>
-        </Fragment>
-      )}
-    </Fragment>
+    <div style={{ height: '1200px' }}>
```
yizheliu-amazon, line 206: This is the only change for full screen issue. The root cause is that not all parent divs have specified 100% height or  explicit px for height. More details can be found here: https://stackoverflow.com/a/31728799

yizheliu-amazon: I manually fix the height for dashboard to 1200 px, which looks appropriate on UI, without too much empty space at the bottom

**Without anomaly**
Screen Shot 2020-09-04 at 11 33 22 AM

**With anomaly**
Screen Shot 2020-09-04 at 11 34 48 AM

public/pages/Dashboard/Components/AnomaliesLiveChart.tsx
ylwu-amzn: Looks weird to set detector name as '' or ' ' , can you explain why we need an empty detector name or space as detector name ?

yizheliu-amazon: When there exists liveAnomalyData, using '' the empty string makes sure those placeholder data won't be shown in legend because its name is ''; When liveAnomalyData does not exist, using space string as detector name can make sure data is plotted in the chart, and because in the case of no liveAnomalyData, legend is hidden, then we are able to see the live chart shown, even without actual anomaly data. Actually the SPACE_STR can be anything, since legend is hidden in that case.

ylwu-amzn (approved):

ohltyler:
Can you confirm if all tests and/or test snapshots are passing?

yizheliu-amazon:
> Can you confirm if all tests and/or test snapshots are passing?

Sure. Run it on my local workspace:
```
Test Suites: 1 skipped, 47 passed, 47 of 48 total
Tests:       4 skipped, 220 passed, 224 total
Snapshots:   35 passed, 35 total
Time:        44.994s
Ran all test suites.
Done in 49.45s

Issue #23, opened 2020-04-15
Improve use of redux across multiple components in the same page


yizheliu-amazon:
Right now we have issue with global redux used in same page but across multiple components. Current work around is to dispatch action correspondingly on those components. But the downside is that we are not able to tell if data is still loading or not in real time. 

This is raised by [comment in PR](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/17#discussion_r406616165).

The potential solution is to leverage `namespace` as per [here](https://stackoverflow.com/a/42939732)

ohltyler:
Another potential solution is to make the reducer functions and corresponding data more granular to allow for retrieving different results from the redux store, rather than needing to call the same function multiple times, and needing to store the results from multiple function calls.

PR #312, opened 2020-10-15
swith direct index query to api call

ylwu-amzn:
*Issues https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/195*
## Description of changes:

Change direct index query to API call for FGAC(Fine-Grained Access Control).

Skip flaky test now, will fix it later.

### DIFF
--- a/public/pages/DetectorResults/containers/AnomalyHistory.tsx
+++ b/public/pages/DetectorResults/containers/AnomalyHistory.tsx
@@ -47,11 +47,11 @@ import { AnomaliesChart } from '../../AnomalyCharts/containers/AnomaliesChart';
 import { FeatureBreakDown } from '../../AnomalyCharts/containers/FeatureBreakDown';
 import { minuteDateFormatter } from '../../utils/helpers';
 import { ANOMALY_HISTORY_TABS } from '../utils/constants';
-import { searchES } from '../../../redux/reducers/elasticsearch';
 import { MIN_IN_MILLI_SECS } from '../../../../server/utils/constants';
 import { INITIAL_ANOMALY_SUMMARY } from '../../AnomalyCharts/utils/constants';
 import { MAX_ANOMALIES } from '../../../utils/constants';
 import { getDetectorResults } from '../../../redux/reducers/anomalyResults';
+import { searchResults } from '../../../redux/reducers/anomalyResults';
 
 interface AnomalyHistoryProps {
   detector: Detector;
@@ -97,26 +97,24 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
       try {
         setIsLoadingAnomalyResults(true);
         const anomalySummaryResult = await dispatch(
-          searchES(
-            getAnomalySummaryQuery(
+        searchResults(getAnomalySummaryQuery(
               dateRange.startDate,
               dateRange.endDate,
               props.detector.id
-            )
-          )
+            ))
         );
         setPureAnomalies(parsePureAnomalies(anomalySummaryResult));
         setBucketizedAnomalySummary(parseAnomalySummary(anomalySummaryResult));
+
         const result = await dispatch(
-          searchES(
-            getBucketizedAnomalyResultsQuery(
+          searchResults(getBucketizedAnomalyResultsQuery(
               dateRange.startDate,
               dateRange.endDate,
               1,
               props.detector.id
-            )
-          )
-        );
+            ))
+          );
+
         setBucketizedAnomalyResults(parseBucketizedAnomalyResults(result));
       } catch (err) {
         console.error(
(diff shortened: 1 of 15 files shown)


yizheliu-amazon (approved):

public/redux/reducers/alerting.ts
```diff
@@ -84,6 +85,18 @@ const reducer = handleActions<Monitors>(
...
+      REQUEST: (state: Monitors): Monitors => ({
+        ...state
+      }),
+      SUCCESS: (state: Monitors, action: APIResponseAction): Monitors => ({
+        ...state
+      }),
+      FAILURE: (state: Monitors, action: APIResponseAction): Monitors => ({
+        ...state
+      }),
+    },
```
ohltyler, line 99: Maybe don't need to store alert info in redux store, but can you at least update the `requesting` and `errorMessage` fields? Can be helpful later on when determining loading states and propagating errors

public/redux/reducers/anomalyResults.ts
```diff
@@ -60,6 +61,18 @@ const reducer = handleActions<Anomalies>(
...
+      REQUEST: (state: Anomalies): Anomalies => ({
+        ...state
+      }),
+      SUCCESS: (state: Anomalies, action: APIResponseAction): Anomalies => ({
+        ...state
+      }),
+      FAILURE: (state: Anomalies): Anomalies => ({
+        ...state
+      }),
+    },
```
ohltyler, line 75: same thing here

public/redux/reducers/alerting.ts
ylwu-amzn: Yes, not sure if this will impact other components using `requesting` and `errorMessage`. Will add todo here and add it later.

ohltyler: Sure, not a blocker. If anything, adding `requesting` and `errorMessage` wouldn't hurt, as it would just make the frontend more accurate.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -97,26 +97,24 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
...
-          )
+            ))
         );
         setPureAnomalies(parsePureAnomalies(anomalySummaryResult));
         setBucketizedAnomalySummary(parseAnomalySummary(anomalySummaryResult));
+
         const result = await dispatch(
-          searchES(
-            getBucketizedAnomalyResultsQuery(
+          searchResults(getBucketizedAnomalyResultsQuery(
```
ohltyler, line 110: minor: I see this was already like this before, but I feel that overall we should try to avoid making direct calls and using the results directly, since it may cause errors to use those results/responses in downstream functions, etc. This is where redux store can help here in storing data and error messages. Not a blocker obviously

server/routes/ad.ts
```diff
@@ -314,6 +315,32 @@ const searchDetector = async (
...
+      { body: requestBody }
+    );
+    return {
+      ok: true,
+      response,
+    };
+  } catch (err) {
+    console.log('Anomaly detector - Unable to search anomaly result', err);
+    if (isIndexNotFoundError(err)) {
+      return { ok: true, response: { totalDetectors: 0, detectors: [] } };
```
ohltyler, line 338: minor: this response doesn't matter since the reducer isn't using the responses right now, but this response doesn't make a lot of sense. Maybe just return empty for now and add TODO

saratvemulapalli (approved):
Overall the changes look good to me.

ohltyler (approved):
Thanks for adding! Just left a few other minor comments.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
yizheliu-amazon: we have to use direct calls because we are calling same API in multiple places of same component, but current redux store does not support. We have an issue #23 for this.

public/redux/reducers/alerting.ts
ylwu-amzn: Added toto here, if any client components using these search APIs needs `requesting` and `errorMessage`, feel free to resolve this todo..

public/pages/DetectorResults/containers/AnomalyHistory.tsx
ohltyler: I see, thanks for clarifying. Another solution may be making reducer functions more fine-grained and/or storing more fine-grained results in the redux store. Will add suggestion to the issue.

server/routes/ad.ts
ylwu-amzn: Good catch, will fix this in next PR.

PR #314, opened 2020-10-16
Add UX support for HC detector

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Add UX support for HC detector.

Preview for HC detector:

HC-Preview_0_-_2020-10-15_17 47 19

HC-Preview_1_-_2020-10-15_17 47 19

DetectorResult for HC detector:

HC_DetectorResult-0_-_2020-10-14_21 02 50
HC_DetectorResult-1_-_2020-10-14_21 02 50
HC_DetectorResult-2_-_2020-10-14_21 02 50

### DIFF
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -40,6 +40,12 @@ import {
   HeatmapCell,
 } from '../containers/AnomalyHeatmapChart';
 import {
+  getAnomalyGradeWording,
+  getConfidenceWording,
+  getFeatureBreakdownWording,
+  getFeatureDataWording,
+} from '../utils/anomalyChartUtils';
+import {
   DATE_PICKER_QUICK_OPTIONS,
   INITIAL_ANOMALY_SUMMARY,
 } from '../utils/constants';
@@ -59,8 +65,6 @@ interface AnomaliesChartProps {
   dateRange: DateRange;
   isLoading: boolean;
   showAlerts?: boolean;
-  anomalyGradeSeriesName: string;
-  confidenceSeriesName: string;
   detectorId: string;
   detectorName: string;
   detector?: Detector;
@@ -213,7 +217,7 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                         unit={props.unit}
                         onHeatmapCellSelected={props.onHeatmapCellSelected}
                       />,
-                      props.showAlerts === undefined || !props.showAlerts
+                      props.showAlerts !== true
                         ? [
                             <EuiSpacer size="m" />,
                             <AnomalyOccurrenceChart
@@ -229,9 +233,13 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                               bucketizedAnomalies={false}
                               anomalySummary={INITIAL_ANOMALY_SUMMARY}
                               isLoading={props.isLoading}
-                              anomalyGradeSeriesName="Anomaly grade"
-                              confidenceSeriesName="Confidence"
-                              showAlerts={false}
+                              anomalyGradeSeriesName={getAnomalyGradeWording(
+                                props.showAlerts
+                              )}
+                              confidenceSeriesName={getConfidenceWording(
+                                props.showAlerts
+                              )}
+                              showAlerts={props.showAlerts}
                               detectorId={props.detectorId}
                               detectorName={props.detectorName}
                               detector={props.detector}
@@ -248,7 +256,9 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                             />,
                             <EuiSpacer size="m" />,
                             <FeatureBreakDown
-                              title="Sample feature breakdown"
+                              title={getFeatureBreakdownWording(
+                                props.showAlerts
+                              )}
                               //@ts-ignore
                               detector={props.newDetector}
                               //@ts-ignore
@@ -259,7 +269,9 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                               isLoading={props.isLoading}
                               //@ts-ignore
                               dateRange={props.zoomRange}
-                              featureDataSeriesName="Sample feature output"
+                              featureDataSeriesName={getFeatureDataWording(
+                                props.showAlerts
+                              )}
                               isHCDetector={props.isHCDetector}
                               selectedHeatmapCell={props.selectedHeatmapCell}
                             />,
@@ -279,8 +291,8 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
               bucketizedAnomalies={props.bucketizedAnomalies}
               anomalySummary={props.anomalySummary}
               isLoading={props.isLoading}
-              anomalyGradeSeriesName="Anomaly grade"
-              confidenceSeriesName="Confidence"
+              anomalyGradeSeriesName={getAnomalyGradeWording(props.showAlerts)}
+              confidenceSeriesName={getConfidenceWording(props.showAlerts)}
               showAlerts={props.showAlerts}
               detectorId={props.detectorId}
               detectorName={props.detectorName}
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -13,67 +13,38 @@
  * permissions and limitations under the License.
  */
 
-import React, { useState, useEffect } from 'react';
-import { useDispatch } from 'react-redux';
-import { DurationInputArg2 } from 'moment';
-import moment from 'moment';
-import { get } from 'lodash';
 import dateMath from '@elastic/datemath';
 import {
-  EuiFlexItem,
   EuiFlexGroup,
-  EuiIcon,
+  EuiFlexItem,
   EuiLoadingChart,
-  EuiStat,
+  EuiSpacer,
   EuiSuperDatePicker,
 } from '@elastic/eui';
-import {
-  Chart,
-  Axis,
-  LineSeries,
-  niceTimeFormatter,
-  Settings,
-  Position,
-  LineAnnotation,
-  AnnotationDomainTypes,
-  RectAnnotation,
-  ScaleType,
-  XYBrushArea,
-} from '@elastic/charts';
-import { useDelayedLoader } from '../../../hooks/useDelayedLoader';
+import { get } from 'lodash';
+import moment, { DurationInputArg2 } from 'moment';
+import React, { useState } from 'react';
 import ContentPanel from '../../../components/ContentPanel/ContentPanel';
+import { useDelayedLoader } from '../../../hooks/useDelayedLoader';
 import {
-  AnomalySummary,
-  Monitor,
-  Detector,
+  Anomalies,
   DateRange,
-  MonitorAlert,
+  Detector,
+  Monitor,
 } from '../../../models/interfaces';
-import {
-  prepareDataForChart,
-  filterWithDateRange,
-} from '../../utils/anomalyResultUtils';
-import { AlertsFlyout } from '../components/AlertsFlyout/AlertsFlyout';
-
+import { generateAnomalyAnnotations } from '../../utils/anomalyResultUtils';
 import { AlertsButton } from '../components/AlertsButton/AlertsButton';
-import { darkModeEnabled } from '../../../utils/kibanaUtils';
+import { AnomalyDetailsChart } from '../containers/AnomalyDetailsChart';
 import {
-  AlertsStat,
-  AnomalyStatWithTooltip,
-} from '../components/AnomaliesStat/AnomalyStat';
+  AnomalyHeatmapChart,
+  HeatmapCell,
+} from '../containers/AnomalyHeatmapChart';
 import {
-  INITIAL_ANOMALY_SUMMARY,
-  CHART_FIELDS,
   DATE_PICKER_QUICK_OPTIONS,
-  ANOMALY_CHART_THEME,
+  INITIAL_ANOMALY_SUMMARY,
 } from '../utils/constants';
-import {
-  convertAlerts,
-  generateAlertAnnotations,
-  getAnomalySummary,
-  disabledHistoryAnnotations,
-} from '../utils/anomalyChartUtils';
-import { searchAlerts } from '../../../redux/reducers/alerting';
+import { AnomalyOccurrenceChart } from './AnomalyOccurrenceChart';
+import { FeatureBreakDown } from './FeatureBreakDown';
 
 interface AnomaliesChartProps {
   onDateRangeChange(
@@ -83,10 +54,8 @@ interface AnomaliesChartProps {
   ): void;
   onZoomRangeChange(startDate: number, endDate: number): void;
   title: string;
-  anomalies: any[];
   bucketizedAnomalies: boolean;
   anomalySummary: any;
-  annotations?: any[];
   dateRange: DateRange;
   isLoading: boolean;
   showAlerts?: boolean;
@@ -99,85 +68,29 @@ interface AnomaliesChartProps {
   unit?: string;
   monitor?: Monitor;
   children: React.ReactNode | React.ReactNode[];
+  isHCDetector?: boolean;
+  detectorCategoryField?: string[];
+  onHeatmapCellSelected?(heatmapCell: HeatmapCell): void;
+  selectedHeatmapCell?: HeatmapCell;
+  newDetector?: Detector;
+  zoomRange?: DateRange;
+  anomaliesResult: Anomalies | undefined;
 }
-export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
-  const dispatch = useDispatch();
-  const [anomalySummary, setAnomalySummary] = useState<AnomalySummary>(
-    INITIAL_ANOMALY_SUMMARY
-  );
-  const [showAlertsFlyout, setShowAlertsFlyout] = useState<boolean>(false);
-  const [alertAnnotations, setAlertAnnotations] = useState<any[]>([]);
-  const [isLoadingAlerts, setIsLoadingAlerts] = useState<boolean>(false);
-  const [totalAlerts, setTotalAlerts] = useState<number | undefined>(undefined);
-  const [alerts, setAlerts] = useState<MonitorAlert[]>([]);
-  const [zoomRange, setZoomRange] = useState<DateRange>({
-    ...props.dateRange,
-  });
-  const [zoomedAnomalies, setZoomedAnomalies] = useState<any[]>([]);
 
+export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
   const [datePickerRange, setDatePickerRange] = useState({
     start: 'now-7d',
     end: 'now',
   });
 
-  useEffect(() => {
-    const anomalies = prepareDataForChart(props.anomalies, zoomRange);
-    setZoomedAnomalies(anomalies);
-    setAnomalySummary(
-      !props.bucketizedAnomalies
-        ? getAnomalySummary(
-            filterWithDateRange(props.anomalies, zoomRange, 'plotTime')
-          )
-        : props.anomalySummary
-    );
-    setTotalAlerts(filterWithDateRange(alerts, zoomRange, 'startTime').length);
-  }, [props.anomalies, zoomRange]);
-
-  const handleZoomRangeChange = (start: number, end: number) => {
-    setZoomRange({
-      startDate: start,
-      endDate: end,
-    });
-    props.onZoomRangeChange(start, end);
-  };
-
-  useEffect(() => {
-    async function getMonitorAlerts(monitorId: string, startDateTime: number, endDateTime: number) {
-      try {
-        setIsLoadingAlerts(true);
-
-        const result = await dispatch(
-          searchAlerts(monitorId, startDateTime, endDateTime)
-        );
-
-        setIsLoadingAlerts(false);
-        setTotalAlerts(get(result, 'data.response.totalAlerts'))
-
-        const monitorAlerts = convertAlerts(result);
-        setAlerts(monitorAlerts);
-        const annotations = generateAlertAnnotations(monitorAlerts);
-        setAlertAnnotations(annotations);
-      } catch (err) {
-        console.error(`Failed to get alerts for monitor ${monitorId}`, err);
-        setIsLoadingAlerts(false);
-      }
-    }
-    if (props.monitor && props.dateRange.startDate) {
-      getMonitorAlerts(props.monitor.id, props.dateRange.startDate, props.dateRange.endDate);
-    }
-  }, [props.monitor, props.dateRange.startDate]);
-
-  const anomalyChartTimeFormatter = niceTimeFormatter([
-    zoomRange.startDate,
-    zoomRange.endDate,
-  ]);
+  const anomalies = get(props.anomaliesResult, 'anomalies', []);
 
   const handleDateRangeChange = (startDate: number, endDate: number) => {
     props.onDateRangeChange(startDate, endDate);
-    handleZoomRangeChange(startDate, endDate);
+    props.onZoomRangeChange(startDate, endDate);
   };
 
-  const showLoader = useDelayedLoader(props.isLoading || isLoadingAlerts);
+  const showLoader = useDelayedLoader(props.isLoading);
 
   const handleDatePickerDateRangeChange = (
     start: string,
@@ -200,7 +113,7 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
             startTime.valueOf() >= props.dateRange.startDate &&
             endTime.valueOf() <= props.dateRange.endDate
           ) {
-            handleZoomRangeChange(startTime.valueOf(), endTime.valueOf());
+            props.onZoomRangeChange(startTime.valueOf(), endTime.valueOf());
           } else {
             handleDateRangeChange(startTime.valueOf(), endTime.valueOf());
           }
@@ -209,9 +122,16 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
     }
   };
 
+  const handleDatePickerRangeChange = (start: number, end: number) => {
+    setDatePickerRange({
+      start: moment(start).format(),
+      end: moment(end).format(),
+    });
+  };
+
   const datePicker = () => (
     <EuiSuperDatePicker
-      isLoading={props.isLoading || isLoadingAlerts}
+      isLoading={props.isLoading}
       start={datePickerRange.start}
       end={datePickerRange.end}
       onTimeChange={({ start, end, isInvalid, isQuickSelection }) => {
@@ -240,6 +160,7 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
     return (
       <EuiFlexGroup>
         <EuiFlexItem style={{ marginRight: '8px' }}>{datePicker()}</EuiFlexItem>
+
         <EuiFlexItem style={{ marginLeft: '0px' }}>
           {setUpAlertsButton()}
         </EuiFlexItem>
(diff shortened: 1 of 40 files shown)


yizheliu-amazon:
will fix UT while reviewing

yizheliu-amazon:
current UT passed. will add more UT for new code.
```
Test Suites: 1 skipped, 51 passed, 51 of 52 total
Tests:       5 skipped, 230 passed, 235 total
Snapshots:   46 passed, 46 total
Time:        30.932s
Ran all test suites.
Done in 31.66s

File: public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx [hunk 1]
```diff
@@ -0,0 +1,388 @@
...
+  ): void;
+  onZoomRangeChange(startDate: number, endDate: number): void;
+  anomalies: any[];
+  bucketizedAnomalies: boolean;
+  anomalySummary: any;
+  dateRange: DateRange;
+  isLoading: boolean;
+  showAlerts?: boolean;
+  anomalyGradeSeriesName: string;
+  confidenceSeriesName: string;
```
ohltyler, line 86: Minor: maybe combine this into a single `isSample` param or something, where the anomaly grade and anomaly confidence title names can be declared and maintained within the AnomalyDetailsChart itself, rather than having to pass in different hardcoded constants from other components

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -0,0 +1,388 @@
...
+          console.error(`Failed to get alerts for monitor ${monitorId}`, err);
+          setIsLoadingAlerts(false);
+        }
+      }
+      if (
+        props.monitor &&
+        props.dateRange.startDate &&
+        // only load alert stats for non HC detector
+        props.isHCDetector != undefined &&
+        !props.isHCDetector
```
ohltyler, line 175: minor: combine into one condition? If `props.isHCDetector` is true, it would fail both of these conditions

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -0,0 +1,388 @@
...
+      ) {
+        getMonitorAlerts(
+          props.monitor.id,
+          props.dateRange.startDate,
+          props.dateRange.endDate
+        );
+      }
+    }, [props.monitor, props.dateRange.startDate]);
```
ohltyler, line 183: why only trigger this on `startDate` and not `endDate`?

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
ohltyler: In fact, looks like `showAlerts` prop already does this - could either reuse this prop to determine the title names and/or rename `showAlerts` to something like `isSample`

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
```diff
@@ -0,0 +1,388 @@
...
+              titleSize="s"
+            />
+          </EuiFlexItem>
+          {props.showAlerts && !props.isHCDetector ? (
+            <EuiFlexItem>
+              <AlertsStat
+                monitor={props.monitor}
+                showAlertsFlyout={() => setShowAlertsFlyout(true)}
+                totalAlerts={totalAlerts}
+                isLoading={props.isLoading}
```
ohltyler, line 258: should `isLoading` here also include `isLoadingAlerts`? For example, change to `isLoading={props.isLoading || isLoadingAlerts}

ohltyler: Actually, looks like all places where `props.isLoading` is called, `isLoadingAlerts` is called as well (except for this one case above). Could maybe add a variable at the top that's combining these loading states into one? For example `const isLoading = props.isLoading || isLoadingAlerts`?

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -0,0 +1,519 @@
...
+        COMBINED_OPTIONS.options[0].value
+      )
+    );
+    const [heatmapData, setHeatmapData] = useState<PlotData[]>(
+      originalHeatmapData
+    );
+
+    const [sortByFeildValue, setSortByFeildValue] = useState(
+      SORT_BY_FIELD_OPTIONS[0].value
+    );
```
ohltyler, line 131: minor: typo in `field`

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -0,0 +1,519 @@
...
+            HEATMAP_X_AXIS_DATE_FORMAT
+          ).valueOf();
+
+          const selectedStartDate =
+            selectedEndDate -
+            getHeatmapCellDateRangeInterval(
+              //@ts-ignore
+              heatmapData[0].x,
+              selectedCellIndices[1]
+            );
```
ohltyler, line 205: seems confusing to be taking an end date and subtracting with an interval, maybe rename `Interval` in `getHeatmapCellDateRangeInterval` to something like `Size` or `Length`

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -0,0 +1,519 @@
...
+          setCurrentViewOptions([option]);
+          const displayTopEntityNum = get(option, 'value');
+          const updateHeatmapPlotData = getAnomaliesHeatmapData(
+            props.anomalies,
+            props.dateRange,
+            sortByFeildValue,
+            displayTopEntityNum
+          );
+          setOriginalHeatmapData(updateHeatmapPlotData);
+          setHeatmapData(updateHeatmapPlotData);
```
ohltyler, line 286: Could you explain why we need both `originalHeatmapData` and `heatmapData`?

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+      value < ANOMALY_HEATMAP_COLORSCALE[i + 1][0]
+    ) {
+      return ANOMALY_HEATMAP_COLORSCALE[i + 1][1];
+    }
+  }
+};
+
+const NUM_CELLS = 20;
+
+export const HEATMAP_X_AXIS_DATE_FORMAT = 'MM-DD HH:mm YYYY';
```
ohltyler, line 196: minor: was this format specified by UX? Seems kind of confusing, and wondering if we can change to make it consistent with the other date formats on the page, in the form of 'MM/DD/YY h:mm A'

public/pages/AnomalyCharts/containers/AnomalyOccurrenceChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+  confidenceSeriesName: string;
+  detectorId: string;
+  detectorName: string;
+  detector?: Detector;
+  detectorInterval?: number;
+  unit?: string;
+  monitor?: Monitor;
+  isHCDetector?: boolean;
+  selectedHeatmapCell?: HeatmapCell;
+}
```
ohltyler, line 50: note: in the future we should look at combining these long lists of props into some common shared interfaces, since it looks like most of the chart components use a lot of the same ones. nothing to change for now but can consider it a tech debt item.

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
```diff
@@ -45,9 +48,68 @@ interface FeatureBreakDownProps {
...
+    } else {
+      return props.annotations;
+    }
+  };
+  const getDateRange = () => {
+    if (props.isHCDetector && props.selectedHeatmapCell) {
+      return props.selectedHeatmapCell.dateRange;
+    }
+    return props.dateRange;
+  };
```
ohltyler, line 112: Duplicate function in `AnomalyOccurrenceChart`. Maybe refactor this function and the other `get*` functions in both `AnomalyOccurrenceChart` and `FeatureBreakDown` into a common helpers file?

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+export const ANOMALY_HEATMAP_COLORSCALE = [
+  [0, '#F2F2F2'],
+  [0.2, '#F2F2F2'],
+  [0.2, '#F7E0B8'],
+  [0.4, '#F7E0B8'],
+  [0.4, '#F2C596'],
+  [0.6, '#F2C596'],
+  [0.6, '#ECA976'],
+  [0.8, '#ECA976'],
+  [0.8, '#E78D5B'],
```
ohltyler, line 163: Regarding the issue of anomalies with a grade < 0.2 not being visible on the heatmap color scale, can we add an interval where any grade greater > 0 is differentiated from a grade of 0?

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+
+  for (let i = 0; i < ANOMALY_HEATMAP_COLORSCALE.length - 1; i++) {
+    if (
+      value >= ANOMALY_HEATMAP_COLORSCALE[i][0] &&
+      value < ANOMALY_HEATMAP_COLORSCALE[i + 1][0]
+    ) {
+      return ANOMALY_HEATMAP_COLORSCALE[i + 1][1];
+    }
+  }
+};
```
ohltyler, line 192: Could we just use a `switch` or cascading `if` statement here to determine the corresponding color? This would eliminate the need for `ANOMALY_HEATMAP_COLORSCALE` altogether, and make it easier to add a different color for any value > 0 to help solve the issue of low anomaly grades having the same color as anomaly grades of 0.

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+  dateRange: DateRange,
+  sortType: AnomalyHeatmapSortType = AnomalyHeatmapSortType.SEVERITY,
+  displayTopNum: number
+): PlotData[] => {
+  const entityAnomaliesMap = getEntityAnomaliesMap(anomalies);
+  if (isEmpty(entityAnomaliesMap)) {
+    // put placeholder data so that heatmap won't look empty
+    for (let i = 0; i < displayTopNum; i++) {
+      const blankStrValue = buildBlankStringWithLength(i);
+      entityAnomaliesMap.set(blankStrValue, []);
```
ohltyler, line 217: why put various length empty strings in the map? is it so a new mapping is created for each entry? is this for showing a default empty heatmap in the case of no anomalies?

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+      zmin: 0,
+      zmax: 1,
+      type: 'heatmap',
+      showscale: false,
+      xgap: 2,
+      ygap: 2,
+      opacity: 1,
+      text: numAnomalyGrades,
+      hovertemplate:
+        '<b>Time</b>: %{x}<br>' +
```
ohltyler, line 278: minor / future improvement: could the hover time show the date range for the selected cell instead? That could help the user understand that each cell is still an aggregation over a given time frame

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -138,3 +150,281 @@ export const disabledHistoryAnnotations = (
...
+    {
+      ...cloneDeep(heatmapData),
+      z: selectedZData,
+      colorscale: [
+        [0, colorForCell],
+        [1, colorForCell],
+      ],
+      opacity: 1,
+      hoverinfo: 'skip',
+      hovertemplate: null,
```
ohltyler, line 427: why is the `hovertemplate` `null` here? Also, when selecting a cell, it still shows the regular hover template with regular data.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -99,85 +68,29 @@ interface AnomaliesChartProps {
...
+  const anomalies = get(props.anomaliesResult, 'anomalies', []);
 
   const handleDateRangeChange = (startDate: number, endDate: number) => {
     props.onDateRangeChange(startDate, endDate);
-    handleZoomRangeChange(startDate, endDate);
+    props.onZoomRangeChange(startDate, endDate);
   };
 
-  const showLoader = useDelayedLoader(props.isLoading || isLoadingAlerts);
+  const showLoader = useDelayedLoader(props.isLoading);
```
ylwu-amzn, line 93: Any impact to single entity detector by removing `isLoadingAlerts `?

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -256,183 +177,129 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
...
+                        showAlerts={props.showAlerts}
+                        monitor={props.monitor}
+                        detectorInterval={props.detectorInterval}
+                        unit={props.unit}
+                        onHeatmapCellSelected={props.onHeatmapCellSelected}
+                      />,
+                      props.showAlerts === undefined || !props.showAlerts
+                        ? [
+                            <EuiSpacer size="m" />,
+                            <AnomalyOccurrenceChart
```
ylwu-amzn, line 219: `AnomalyOccurrenceChart` is only used by HC detector and `AnomalyDetailChart` is only used single entity detector?

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -256,183 +177,129 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
...
+                              bucketizedAnomalies={false}
+                              anomalySummary={INITIAL_ANOMALY_SUMMARY}
+                              isLoading={props.isLoading}
+                              anomalyGradeSeriesName="Anomaly grade"
+                              confidenceSeriesName="Confidence"
+                              showAlerts={false}
+                              detectorId={props.detectorId}
+                              detectorName={props.detectorName}
+                              detector={props.detector}
+                              detectorInterval={get(
```
ylwu-amzn, line 238: We have passed `detector`, why still pass `detectorInterval` ? We can get `detectorInterval` from detector.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -123,18 +139,18 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
         );
       } finally {
         setIsLoadingAnomalyResults(false);
-        fetchRawAnomalyResults(false);
       }
     }
 
+    fetchRawAnomalyResults(isHCDetector);
```
ohltyler, line 145: according to `fetchRawAnomalyResults()`, the param is related to loading or not - why does `isHCDetector` impact this?

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -266,7 +268,8 @@ export function AnomalyResults(props: AnomalyResultsProps) {
       return get(
         getFeatureDataMissingMessageAndActionItem(
           featureMissingSeverity,
-          featureNamesAtHighSev
+          featureNamesAtHighSev,
+          !isEmpty(detector.categoryField)
```
ohltyler, line 272: change to `isHCDetector`

public/pages/DetectorResults/containers/AnomalyResults.tsx
```diff
@@ -354,7 +358,8 @@ export function AnomalyResults(props: AnomalyResultsProps) {
         {get(
           getFeatureDataMissingMessageAndActionItem(
             featureMissingSeverity,
-            featureNamesAtHighSev
+            featureNamesAtHighSev,
+            !isEmpty(detector.categoryField)
```
ohltyler, line 362: change to `isHCDetector`

public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
```diff
@@ -51,9 +57,9 @@ export function AnomalyResultsTable(props: AnomalyResultsTableProps) {
...
     },
   });
-  const [targetAnomalies, setTargetAnomalies] = useState<AnomalyData[]>([]);
+  const [targetAnomalies, setTargetAnomalies] = useState<any[]>([] as any[]);
   const totalAnomalies = props.anomalies
-    ? props.anomalies.filter(anomaly => anomaly.anomalyGrade > 0)
+    ? props.anomalies.filter((anomaly) => anomaly.anomalyGrade > 0)
     : [];
 
   const sortFiledCompare = (field: string, sortDirection: SORT_DIRECTION) => {
```
ohltyler, line 65: minor: typo in `field`

public/pages/DetectorResults/containers/AnomalyResultsTable.tsx
```diff
@@ -68,7 +74,14 @@ export function AnomalyResultsTable(props: AnomalyResultsTableProps) {
 
   useEffect(() => {
     const anomalies = props.anomalies
-      ? props.anomalies.filter(anomaly => anomaly.anomalyGrade > 0)
+      ? props.anomalies
+          .filter((anomaly) => anomaly.anomalyGrade > 0)
+          .map((anomaly) => {
+            return {
+              ...anomaly,
+              [ENTITY_VALUE_FIELD]: get(anomaly, 'entity[0].value'),
```
ohltyler, line 82: To save space, maybe only add this extra entity field if `isHCDetector` flag is `true`? I'm thinking of the case where there may be a lot of anomalies for a non-HC detector where this field is unnecessary

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -352,49 +362,49 @@ export const getBucketizedAnomalyResultsQuery = (
...
-                ],
-                _source: {
-                  includes: RETURNED_AD_RESULT_FIELDS,
                 },
-                size: 1,
+              ],
+              _source: {
+                includes: RETURNED_AD_RESULT_FIELDS,
               },
+              size: 1,
```
ylwu-amzn, line 402: If try to get last 7 days AD result for HC detector, this query will return only the top 1 result even user choose top 10 ?

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -144,10 +160,8 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
     1
   );
 
-  const fetchRawAnomalyResults = async (
-    shouldSetAtomicAnomalyResults: boolean
-  ) => {
-    if (shouldSetAtomicAnomalyResults) {
+  const fetchRawAnomalyResults = async (showLoading: boolean) => {
```
ylwu-amzn, line 163: As we discussed, get raw anomaly result may have performance issue if user choose a long period for a HC detector which has thousands of entities. Suppose we get 1K result for 1 entity, 1K entities will have 1M results.
Add some comments & TODO here and fix it later.

public/pages/EditFeatures/containers/SampleAnomalies.tsx
```diff
@@ -74,6 +76,10 @@ export function SampleAnomalies(props: SampleAnomaliesProps) {
     endDate: initialEndDate.valueOf(),
   });
 
+  const [selectedHeatmapCell, setSelectedHeatmapCell] = useState<HeatmapCell>();
+
+  const isHCDetector = !isEmpty(get(newDetector, 'categoryField', []));
```
ohltyler, line 81: minor: is adding the `categoryFields` prop necessary if it can already be retrieved via `detector` prop?

public/pages/DetectorResults/containers/AnomalyHistory.tsx
ylwu-amzn: Delete this, comment wrong method

public/pages/EditFeatures/containers/SampleAnomalies.tsx
```diff
@@ -135,13 +145,16 @@ export function SampleAnomalies(props: SampleAnomaliesProps) {
...
+    setSelectedHeatmapCell(undefined);
     try {
       const updatedDetector = prepareDetector(
         props.featureList,
         props.shingleSize,
+        props.categoryFields,
         newDetector,
         true
       );
+      console.log('updatedDetector in SampleAnomalies', updatedDetector);
```
ohltyler, line 157: remove old log

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
yizheliu-amazon: sure. good idea. I agree for now we can just reuse `showAlerts` to do that.

public/redux/reducers/anomalies.ts
```diff
@@ -24,12 +24,12 @@ import { Anomalies } from '../../models/interfaces';
 
 const PREVIEW_DETECTOR = 'ad/PREVIEW_DETECTOR';
 
-export interface Anomalies {
+export interface PreviewAnomalies {
```
ohltyler, line 27: Minor: may want to change the filename completely to `previewAnomalies.ts` to keep it consistent

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
yizheliu-amazon: refactored this.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
ohltyler: same with id and name

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
yizheliu-amazon: good catch. we should also show alert stats when `props.isHCDetector` is undefined here. As long as non-HC detector, we show alert. Changing to `props.isHCDetector !== true`

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
yizheliu-amazon: good catch. previous`getMonitorAlerts()` only accepts startDate, now it accepts start/end date now. Changing to include endDate

public/pages/AnomalyCharts/containers/AnomalyDetailsChart.tsx
yizheliu-amazon: good point. I will change to `const isLoading = props.isLoading || isLoadingAlerts` for clean code.

public/pages/AnomalyCharts/containers/AnomalyOccurrenceChart.tsx
yizheliu-amazon: sure. created issue for this: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/316

public/pages/AnomalyCharts/containers/FeatureBreakDown.tsx
yizheliu-amazon: sure. will add it.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: good catch. fixing

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: agree. changing to `Length`.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: `originalHeatmapData` has data for each combined option, and `heatmapData` is only for data rendered on Heatmap. For example, if `top 20` is selected, both `originalHeatmapData` and `heatmapData` has same top 20 entities data. Then user selects 1 individual entity A, then `heatmapData` only has data for entity A, while `originalHeatmapData` still has top 20 entities data. We need `originalHeatmapData` to get individual entity options.

ohltyler: i see, makes sense. Thanks

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
yizheliu-amazon: if not set to null, it will have extra chevron as there are 2 traces on the heatmap
extra-chevron

ohltyler: I see yeah, remember this issue. Cool thanks.

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
yizheliu-amazon: I think using `switch` or cascading `if` may not be flexible when a different color/value is added. With current implementation, we don't need to make any change for this method if a new color/value is added. I will add comment to explain saying 1st `if ` is to check if value is larger than biggest value in color scale, 2nd `if` is to check value is smaller than smallest value. 

And the issue of low anomaly grades having 0 grade color is fixed by adding 2 lines above, to specify for number very very close to 0, give it color 0 grade color. This is some tricky way as specified [here](https://plotly.com/javascript/colorscales/#custom-discretized-heatmap-colorscale)
```
  [0.0000001, '#F2F2F2'],
  [0.0000001, '#F7E0B8'],

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 8 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:198]: Good point. This is actually more for the purpose of getting date range interval, with no `Year` info, there may be issue when [calculating time difference between 2 neighbor timestamps](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/314/files#diff-a044985587be3c93885ad08b0bb90396e88600895b94ace004c30cda74f955d7R233-R234) in 1 special case, where previous timestamp is 12/31, 23:59, and the next one is 01/01, 00:01. In that case, time difference from 01/01, 00:01 to 12/31, 23:59 is totally different from 12/31, 23:59 to 01/01, 00:01. I think I will just add time interval as 1 attribute of PlotData so that we can get rid of `getHeatmapCellDateRangeInterval()`

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 12 above)
Comment (ohltyler) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:194]: I see, yeah that's fair enough. I think adding those 2 lines is a good solution

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 13 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:219]: yeah. It is just to show default empty heatmap in the case of no anomalies. kind of hacky. If I only put 10 same length whitespace string as entity value, aka Y values, it will show like this:
Screen Shot 2020-10-16 at 5 57 58 PM

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 14 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:280]: It is hard to implement that. I didn;t find solution. Original UX has such design to show time range, but due to limitation of Plotly library, we can only have 4 fields(x, y, z, text) of `PlotData` of each cell shown in [hovertemplate](https://plotly.com/javascript/reference/heatmap/#heatmap-hovertemplate). And `x` must be timestamp of string, there is no good way to customize/format any value to date string for `x`.

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 13 above)
Comment (ohltyler) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:219]: yeah, makes sense. could you maybe add a comment describing why you had to do this though?

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 14 above)
Comment (ohltyler) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:280]: Yeah, i see, not a problem, just wanted to confirm there was no easy way.

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 8 above)
Comment (ohltyler) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:198]: Sorry, not sure I understand. Wouldn't adding the year in the format of 'MM/DD/YY h:mm A' solve this, by comparing against the year as well? To prevent issues with that edge case you mentioned above?

File: public/pages/AnomalyCharts/containers/AnomaliesChart.tsx (hunk 16 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/containers/AnomaliesChart.tsx:93]: no. it will be there for `AnomalyDetailsChart`

File: public/pages/AnomalyCharts/containers/AnomaliesChart.tsx (hunk 17 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/containers/AnomaliesChart.tsx:219]: Yes, and `AnomalyOccurrenceChart ` contains `AnomalyDetailChart`. `AnomalyOccurrenceChart` is actually wrapper of `AnomalyDetailChart`, making `AnomalyDetailChart` a separate chart so that it can be easily re-used in multiple places

File: public/pages/DetectorResults/containers/AnomalyHistory.tsx (hunk 19 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyHistory.tsx:145]: For HC detector, we are relying on raw data to get data for heatmap, that is why we want to show the loading spinner; For single entity detector, fetching raw data should not impact loading state. If fetching large time range data, we use `isLoadingAnomalyResults` to control loading spinner; if fetching small time range data, it is every quick for single entity detector, I think skipping loading spinner is fine for small time range case. I can change to even show loading spinner for small time range in case of single entity detector, it just added a few more `if - else`. Please feel free to let me know your thoughts.

File: public/pages/DetectorResults/containers/AnomalyResults.tsx (hunk 20 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyResults.tsx:272]: good catch. will change it.

File: public/pages/DetectorResults/containers/AnomalyResults.tsx (hunk 21 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyResults.tsx:362]: sure.

File: public/pages/DetectorResults/containers/AnomalyResultsTable.tsx (hunk 22 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyResultsTable.tsx:65]: thanks for catching.

File: public/pages/EditFeatures/containers/SampleAnomalies.tsx (hunk 26 above)
Comment (yizheliu-amazon) [reply public/pages/EditFeatures/containers/SampleAnomalies.tsx:81]: if we get it from `detector` prop, it will always be false in case of single entity detector. 

If we retrieve it from `SampleAnomaliesProps`, when user already views Preview Result chart without enabling category field, the Preview Result Chart below will be automatically changed from line chart to heatmap when user enables/selects category field. 

In current implementation, in case of user already views Preview Result chart as single detector, only after user enables/selects category field, and clicks "Refresh", then result chart is changed to heatmap. This provides better UX.

File: public/pages/EditFeatures/containers/SampleAnomalies.tsx (hunk 27 above)
Comment (yizheliu-amazon) [reply public/pages/EditFeatures/containers/SampleAnomalies.tsx:157]: good catching. removing

File: public/pages/utils/anomalyResultUtils.ts (hunk 24 above)
Comment (yizheliu-amazon) [reply public/pages/utils/anomalyResultUtils.ts:402]: Yes. Actually right now we are relying on raw data for HC detector. I think we should be able to disable getting bucketized data since we don't rely on it. will change it.

File: public/redux/reducers/anomalies.ts (hunk 28 above)
Comment (yizheliu-amazon) [reply public/redux/reducers/anomalies.ts:27]: will try to change it.

File: public/pages/AnomalyCharts/containers/AnomaliesChart.tsx (hunk 18 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/containers/AnomaliesChart.tsx:238]: Good catch. will change it.

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 8 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:198]: sorry it is confusing. Actually format in current single entity anomaly chart has no year in X axis. Using 'MM/DD/YY h:mm A' may not keep it consistent. But adding 1 more attribute solves the problem in a clean way. will push new commit after addressing all comments.

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts (hunk 13 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/utils/anomalyChartUtils.ts:219]: sure. added comment.

File: public/pages/DetectorResults/containers/AnomalyResultsTable.tsx (hunk 23 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyResultsTable.tsx:82]: sure. will add such logic only for HC detector

File: public/pages/DetectorResults/containers/AnomalyHistory.tsx (hunk 19 above)
Comment (ohltyler) [reply public/pages/DetectorResults/containers/AnomalyHistory.tsx:145]: I see, makes sense, thanks for explaining. I think it is good as is - I haven't noticed loading issues while testing.

File: public/pages/EditFeatures/containers/SampleAnomalies.tsx (hunk 26 above)
Comment (ohltyler) [reply public/pages/EditFeatures/containers/SampleAnomalies.tsx:81]: I see, this makes sense in the context of the sample anomalies page as part of the edit features page.

File: public/pages/AnomalyCharts/containers/AnomaliesChart.tsx (hunk 18 above)
Comment (yizheliu-amazon) [reply public/pages/AnomalyCharts/containers/AnomaliesChart.tsx:238]: refactored

File: public/pages/DetectorResults/containers/AnomalyResultsTable.tsx (hunk 23 above)
Comment (yizheliu-amazon) [reply public/pages/DetectorResults/containers/AnomalyResultsTable.tsx:82]: changed.

File: public/pages/AnomalyCharts/utils/anomalyChartUtils.ts [hunk 29]
```diff
@@ -232,6 +236,24 @@ export const getAnomaliesHeatmapData = (
...
+          entityAnomalies.filter(
+            (anomaly) => get(anomaly, 'anomalyGrade', 0) > 0
+          )
+        )) &&
+      !isEmpty(entity.trim())
+    ) {
+      console.log(
+        `find entity ${entity} with empty anomalies`,
+        entityAnomalies
+      );
```
ohltyler: minor: remove this log

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -430,3 +453,50 @@ export const getSelectedHeatmapCellPlotData = (
     },
   ] as PlotData[];
 };
+
+export const getAnomalyGradeWording = (isNotSample: boolean | undefined) => {
+  return isNotSample ? 'Anomaly grade' : 'Sample anomaly grade';
+};
```
ohltyler, line 462: minor: usually prefer the default booleans to have a positive connotation rather than a negative one - like changing `isNotSample` to `isSample`. Not a blocker

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -430,3 +453,50 @@ export const getSelectedHeatmapCellPlotData = (
...
+
+export const getConfidenceWording = (isNotSample: boolean | undefined) => {
```
ohltyler, line 464: same here

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
ohltyler: others like that below as well, feel free to ignore

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -153,10 +149,14 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
...
+      detectorId={get(props.detector, 'id', '')}
+      detectorName={get(props.detector, 'name', '')}
+      detectorInterval={get(
+        props.detector,
+        'detectionInterval.period.interval',
+        1
+      )}
+      unit={get(props.detector, 'detectionInterval.period.unit', 'Minutes')}
     />
   );
```
ohltyler, line 161: any particular reason why you didn't continue the removal of props to `AlertsButton` and `AnomalyHeatmapChart`?

public/pages/EditFeatures/containers/SampleAnomalies.tsx
```diff
@@ -235,8 +235,9 @@ export function SampleAnomalies(props: SampleAnomaliesProps) {
                 onZoomRangeChange={handleZoomChange}
                 isLoading={isLoading}
                 dateRange={dateRange}
-                detectorId={props.detector.id}
-                detectorName={props.detector.name}
+                // detectorId={props.detector.id}
+                // detectorName={props.detector.name}
+                detector={props.detector}
```
ohltyler, line 240: minor: can we remove these commented out lines altogether?

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
yizheliu-amazon: alert button is still needed. We need it to redirect user to Alerting page, but we won't show alerts info, as the alerts related UX is not finalized yet for HC detector.

public/pages/EditFeatures/containers/SampleAnomalies.tsx
yizheliu-amazon: good catch. will remove them.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
ohltyler: Sorry, I meant why you didn't also refactor the `AlertsButton` and `AnomalyHeatmapChart` detector props into a single `detector` prop as well.

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
yizheliu-amazon: In existing code, `props.showAlerts` == true, then show wording **without** `Sample`, and if it is false or not specified, we show something like `Sample anomaly grade`

We might have to keep using this unless we do more refactoring.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
yizheliu-amazon: oh, I see. I think it is not very necessary to pass a whole detector info to it because AlertsButton needs only a few detector info, while `detector` has other info such as state, category field as well.

ohltyler: That's fair, makes sense.

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
ohltyler: Yeah sure, seems fine, just thought I'd throw that out there.

ohltyler (approved):
LGTM!

ylwu-amzn (approved):

PR #320, opened 2020-10-20
Show user friendly message when permission error arises

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Show user friendly message when permission error arises

Example error log of kibana backend:

```
Anomaly detector - PutDetector { Error: [security_exception] no permissions for [cluster:admin/opendistro/ad/detector/write] and User [name=bob, backend_roles=[odfe], requestedTenant=null]
    at respond (/local/apollo/var/env/SearchServicesHome/home/yizheliu/workspace/kibana-791/kibana/node_modules/elasticsearch/src/lib/transport.js:349:15)
    at checkRespForFailure (/local/apollo/var/env/SearchServicesHome/home/yizheliu/workspace/kibana-791/kibana/node_modules/elasticsearch/src/lib/transport.js:306:7)
    at HttpConnector.<anonymous> (/local/apollo/var/env/SearchServicesHome/home/yizheliu/workspace/kibana-791/kibana/node_modules/elasticsearch/src/lib/connectors/http.js:173:7)
    at IncomingMessage.wrapper (/local/apollo/var/env/SearchServicesHome/home/yizheliu/workspace/kibana-791/kibana/node_modules/lodash/lodash.js:4949:19)
    at IncomingMessage.emit (events.js:203:15)
    at endReadableNT (_stream_readable.js:1145:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
  status: 403,
  displayName: 'AuthorizationException',
  message:
   '[security_exception] no permissions for [cluster:admin/opendistro/ad/detector/write] and User [name=bob, backend_roles=[odfe], requestedTenant=null]',
  path: '/_opendistro/_anomaly_detection/detectors',
  query: {},
  body:
   { error:
      { root_cause: [Array],
        type: 'security_exception',
        reason:
         'no permissions for [cluster:admin/opendistro/ad/detector/write] and User [name=bob, backend_roles=[odfe], requestedTenant=null]' },
     status: 403 },
  statusCode: 403,
  response:
   '{"error":{"root_cause":[{"type":"security_exception","reason":"no permissions for [cluster:admin/opendistro/ad/detector/write] and User [name=bob, backend_roles=[odfe], requestedTenant=null]"}],"type":"security_exception","reason":"no permissions for [cluster:admin/opendistro/ad/detector/write] and User [name=bob, backend_roles=[odfe], requestedTenant=null]"},"status":403}',
  toString: [Function],
  toJSON: [Function] }
```

Before:
Screen Shot 2020-10-20 at 3 28 04 AM

After
Screen Shot 2020-10-20 at 1 55 38 AM
Screen Shot 2020-10-20 at 1 55 14 AM

### DIFF
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -35,7 +35,6 @@ import { SORT_DIRECTION, AD_DOC_FIELDS } from '../utils/constants';
 import {
   mapKeysDeep,
   toCamel,
-  toSnake,
   toFixedNumberForAnomaly,
 } from '../utils/helpers';
 import {
@@ -48,8 +47,9 @@ import {
   getDetectorsWithJob,
   getDetectorInitProgress,
   isIndexNotFoundError,
+  getErrorMessage,
 } from './utils/adHelpers';
-import { set } from 'lodash';
+import { isEmpty, set } from 'lodash';
 
 type PutDetectorParams = {
   detectorId: string;
@@ -89,7 +89,10 @@ const deleteDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - deleteDetector', err);
-    return { ok: false, error: err.body || err.message };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
@@ -116,7 +119,10 @@ const previewDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - previewDetector', err);
-    return { ok: false, error: err };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
@@ -160,11 +166,10 @@ const putDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - PutDetector', err);
-    //FIXME: This is temporary as backend should send error message inside message instead of body
-    if (err.statusCode === 400) {
-      return { ok: false, error: err.body.error.reason };
-    }
-    return { ok: false, error: err.message.data.error };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
@@ -215,7 +220,10 @@ const getDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - Unable to get detector', err);
-    return { ok: false, error: err.message };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
@@ -235,7 +243,10 @@ const startDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - startDetector', err);
-    return { ok: false, error: err.body || err.message };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
@@ -255,7 +266,10 @@ const stopDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - stopDetector', err);
-    return { ok: false, error: err.body || err.message };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
   }
 };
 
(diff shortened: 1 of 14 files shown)


server/routes/ad.ts
```diff
@@ -89,7 +89,10 @@ const deleteDetector = async (
     };
   } catch (err) {
     console.log('Anomaly detector - deleteDetector', err);
-    return { ok: false, error: err.body || err.message };
+    return {
+      ok: false,
+      error: getErrorMessage(err),
+    };
```
ohltyler, line 95: Since `getErrorMessage()` doesn't check for `err.body` at all, seems like this will change the functionality. Can you confirm that the error messages returned here make sense?

server/sampleData/utils/helpers.ts
```diff
@@ -104,10 +106,12 @@ export const loadSampleData = (
...
         if (resp.errors) {
-          console.log('Error while bulk inserting. ', resp.errors);
-          return Promise.reject(
-            new Error('Error while bulk inserting. Please try again.')
-          );
+          const errorItems = resp.items;
+          const firstErrorReason = isEmpty(errorItems)
+            ? 'Error while bulk inserting. Please try again.'
+            : prettifyErrorMessage(errorItems[0].index.error.reason);
+          console.log(firstErrorReason);
```
ohltyler, line 113: minor: change to `console.error` or remove

server/utils/helpers.ts
```diff
@@ -62,3 +63,19 @@ export const formatAnomalyNumber = (num: number): string => {
...
+export const prettifyErrorMessage = (rawErrorMessage: string) => {
+  if (isEmpty(rawErrorMessage)) {
+    return 'Unknow error is returned.';
+  }
+  const match = rawErrorMessage.match(PERMISSIONS_ERROR_PATTERN);
+  if (isEmpty(match)) {
+    return rawErrorMessage;
+  }
+  //@ts-ignore
+  return `User ${match[2]} has no permissions to [${match[1]}].`;
```
ohltyler, line 80: minor: if we add the bottom message in an `else` block can we remove the `//@ts-ignore`, since `match` won't be empty?

server/routes/ad.ts
yizheliu-amazon: It is basically same as before. For `err.body || err.message`, if `body` is not there, it will return `message` instead. For `getErrorMessage()` we try to get  `err.body.error.reason`, if `body` is undefined, we change to `message` as well.

The only case is `err.body` is there, but `err.body.error` is undefined. But I have double checked exception log and all of APIs actually have both `err.body.error.reason` and `message`. It should be good for us.

server/sampleData/utils/helpers.ts
yizheliu-amazon: sure. good catch. will remove it.

server/utils/helpers.ts
yizheliu-amazon: Sure. I can change to that.

server/routes/ad.ts
ohltyler: Got it, yeah I was thinking that maybe if `err.body` was there, then changing to just use `getErrorMessage()` which only uses `err.message` then we would possibly return a different message. Thanks for checking the apis. Seems that the backend error handling was probably less standardized when first implemented

yizheliu-amazon: yeah. It might also be caused by our AD ES plugin exception format is not standardized somehow.

ohltyler (approved):
LGTM, thanks for adding!

ylwu-amzn (approved):
LGTM ,thanks for the quick fix

saratvemulapalli (approved):
Thanks for taking care of this.
This looks great.

PR #322, opened 2020-10-22
Query anomaly only for HC detector

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Query anomaly only for HC detector. This is for issue found during performance test: in current implementation, if there are over 10K(our upper limit) anomaly results during query time range, we only get 10K results, including both 0 and >0 result, so that we may miss anomalies with grade > 0. It is easy to get into such case for HC detector. 

And actually HC detector for now only uses anomalyResults with grade > 0, thus I change to only get anomalyResults with grade > 0 for HC detector to avoid above case.


ohltyler (approved):
LGTM, thanks for the change

weicongs-amazon (approved):

ylwu-amzn (approved):

PR #324, opened 2020-10-22
Add match and count APIs

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR adds frontend support for the implemented `match` and `count` APIs (see [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/286)).

These apis are used in 2 different parts for validation on the create/edit detector page, replacing the search API which caused a security leak by returning extra detector information for a user which they may not be supposed to have access to.

Notes:
- confirmed no UT breaks
- confirmed both APIs return correct information
- tested on edit and create to confirm the validation for duplicate name worked appropriately

### DIFF
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -71,6 +71,8 @@ export default function (apiRouter: Router) {
   apiRouter.post('/detectors/{detectorId}/start', startDetector);
   apiRouter.post('/detectors/{detectorId}/stop', stopDetector);
   apiRouter.get('/detectors/{detectorId}/_profile', getDetectorProfile);
+  apiRouter.get('/detectors/{detectorName}/_match', matchDetector);
+  apiRouter.get('/detectors/_count', getDetectorCount);
 }
 
 const deleteDetector = async (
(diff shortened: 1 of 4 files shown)


ylwu-amzn (approved):

server/routes/ad.ts
```diff
@@ -735,6 +737,45 @@ const getAnomalyResults = async (
...
 };
 
+const matchDetector = async (
+  req: Request,
+  h: ResponseToolkit,
+  callWithRequest: CallClusterWithRequest
+): Promise<ServerResponse<AnomalyResults>> => {
+  try {
+    const { detectorName } = req.params;
+    console.log('calling matchDetector() w/ detector name: ', detectorName);
```
yizheliu-amazon, line 747: extra log

server/routes/ad.ts
```diff
@@ -735,6 +737,45 @@ const getAnomalyResults = async (
   }
 };
 
+const matchDetector = async (
```
yizheliu-amazon, line 740: just curious: what does this API do? search detector by name?

server/routes/ad.ts
```diff
@@ -735,6 +737,45 @@ const getAnomalyResults = async (
...
+    const response = await callWithRequest(req, 'ad.matchDetector', {
+      detectorName,
+    });
+    return {
+      ok: true,
+      response: response,
+    };
+  } catch (err) {
+    console.log('Anomaly detector - matchDetector', err);
+    return { ok: false, error: err.message };
```
yizheliu-amazon, line 757: maybe use `getErrorMessage()` ?

server/routes/ad.ts
ohltyler: thanks, removed from `getDetectorCount()` too

yizheliu-amazon:
can you quickly check if e2e test passes?

server/routes/ad.ts
ohltyler: good idea, will change

yizheliu-amazon (approved):

server/routes/ad.ts
ohltyler: basically returns `true` or `false` if the passed params match any existing detector name, without returning any extra detector information that a user may not have access to if using FGAC

ohltyler: See backend PR for more https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/286

ohltyler:
> can you quickly check if e2e test passes?

Getting a flaky test failing on my local, but these changes shouldn't affect it, I imagine it should still pass fine in the github runner

saratvemulapalli (approved):
Thanks Tyler for the change!

PR #328, opened 2020-11-18
Migrate entire plugin to new platform

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR migrates the entire plugin to the new Kibana platform, which is required for Kibana 7.10. A follow up PR will be made to upgrade the plugin to 7.10. Overview of the changes can be found [here](https://github.com/elastic/kibana/blob/master/src/core/MIGRATION.md#overview).

**The main changes include:**
- Defining separate lifecycles for browser-side plugin and server-side plugin by having separate `index` and `plugin` files for each
- Changing the overall entry point of the plugin to be in the new `kibana.json` manifest file
- Migrating the browser client from Angular's `IHttpService` to Kibana's core `HttpSetup` client, and changing all request/response parsing logic accordingly
- Changing the server's communication with the backend by registering an `IClusterClient` with all of the services, and changing the different route functions to be grouped into different service classes
- Change in accessing core services: removing all legacy import `ui/*` statements, and accessing core services (chrome breadcrumbs, toast notifications, etc.) by consuming them via React context
- Complete removal of all dependencies on Angular
- Updating all unit tests and adding a core services mock to be used for components that are consuming core services

**Notes:**
- All unit tests pass
- All integration tests pass
- Sanity tested most plugin functionality
- Sanity tested most server/node APIs and most backend API calls
- Checked dark mode
- Cannot run the plugin via `yarn start` in the plugin repo anymore - need to run in base Kibana dir

**Known issues:**
- `yarn build` issues - the built artifact won't work with the binary Kibana. This is fixed separately, will be included in a follow up PR

**Task backlog:**
- Sanity test w/ alerting when it's finished
- Sanity test w/ FGAC enabled
- Update the README with any necessary `yarn` command changes


public/anomaly_detection_app.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Provider } from 'react-redux';
+import configureStore from './redux/configureStore';
+import { CoreServicesContext } from './components/CoreServices/CoreServices';
+
+export function renderApp(coreStart: CoreStart, params: AppMountParameters) {
+  const http = coreStart.http;
+  const store = configureStore(http);
+
+  // Load Chart's dark mode CSS (if applicable)
+  const isDarkMode = coreStart.uiSettings.get('theme:darkMode') || false;
```
yizheliu-amazon, line 30: can we re-use `darkModeEnabled()` here?

yizheliu-amazon (approved):
thanks for the change.

public/anomaly_detection_app.tsx
ohltyler: I don't think so, since this is before the core services context is consumed by any of the child components, which is how `darkModeEnabled()` works.

Also, this logic is copied over from before.

package.json
```diff
@@ -21,7 +21,7 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run",
     "cy:run-with-security": "cypress run --env SECURITY_ENABLED=true",
-    "test:e2e": "CHOKIDAR_USEPOLLING=1 CYPRESS_responseTimeout=180000 WAIT_ON_TIMEOUT=900000 start-server-and-test 'yarn start-cy-server' http-get://localhost:5601/app/opendistro-anomaly-detection-kibana 'echo sleeping to wait for server to get ready && sleep 180 && yarn cy:run'"
+    "test:e2e": "CHOKIDAR_USEPOLLING=1 CYPRESS_responseTimeout=180000 WAIT_ON_TIMEOUT=900000 start-server-and-test 'cd ../../ && yarn start --no-base-path --oss && cd plugins/anomaly-detection-kibana-plugin' http-get://localhost:5601/app/opendistro-anomaly-detection-kibana 'echo sleeping to wait for server to get ready && sleep 180 && yarn cy:run'"
```
yizheliu-amazon, line 24: minor: `yarn start-cy-server` === `yarn start --no-base-path --oss`

Either removing `start-cy-server` or keeping `yarn start-cy-server` is fine for me. just don't want to keep unused code.

yizheliu-amazon: nvm. I think we can just remove `start-cy-server` script.

ohltyler: Good catch, forgot that I had changed that while testing. I'll go ahead and remove the script

public/redux/reducers/__tests__/anomalyResults.test.ts
```diff
@@ -44,16 +44,19 @@ describe('elasticsearch reducer actions', () => {
...
-        "featureData": undefined,
+        featureData: undefined,
       });
-      expect(httpMockedClient.get).toHaveBeenCalledWith(
+      expect(
+        httpMockedClient.get
+      ).toHaveBeenCalledWith(
         `..${AD_NODE_API.DETECTOR}/${tempDetectorId}/results`,
-        { params: queryParams }
+        { query: JSON.stringify(queryParams) }
```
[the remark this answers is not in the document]
ohltyler: Good catch, in fact the reason the test was written this way was because the reducer is `stringify`ing the query, which isn't necessary. I've removed that altogether and kept as an object, and updated the tests. Confirmed it's still able to pull results correctly.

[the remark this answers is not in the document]
ohltyler: Confirmed no other places are `stringify`ing a query. I think this was left over from debugging the original request parsing issues.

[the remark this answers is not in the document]
ftianli-amzn: Oh, got it. I didn't aware of the reducer. Thanks for testing it!

kibana.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "opendistroAnomalyDetectionKibana",
+  "version": "1.11.0.0",
+  "kibanaVersion": "7.9.1",
```
ylwu-amzn, line 4: Should we change to 7.10?

kibana.json
```diff
@@ -0,0 +1,10 @@
...
+  "configPath": ["opendistro_anomaly_detection_kibana"],
+  "requiredPlugins": ["navigation"],
+  "optionalPlugins": ["opendistro-alerting", "opendistro_alerting"],
```
ylwu-amzn, line 7: Why add two alerting plugins?

kibana.json
ftianli-amzn: This PR only did the job for migration, without the version upgrade. I think keep the Kibana version during migration is easier for testing.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
```diff
@@ -198,11 +197,11 @@ export const DetectorDetail = (props: DetectorDetailProps) => {
   const handleStartAdJob = async (detectorId: string) => {
     try {
       await dispatch(startDetector(detectorId));
-      toastNotifications.addSuccess(
+      core.notifications.toasts.addSuccess(
```
ylwu-amzn, line 200: Can we use `const { httpClient, notifications } = this.props;` to get notifications like alerting Kibana change? The import path of `CoreStart` looks very long `'../../../../../../src/core/public'`

kibana.json
ohltyler: This was the naming conventions for the ES and Kibana alerting plugins. With the new naming conventions for the Kibana plugin this will need to be revisited, I'll make a note of it. Also, adding optional plugins doesn't really have much functionality at all, and we could remove altogether. See [here](https://github.com/elastic/kibana/blob/master/docs/development/core/server/kibana-plugin-core-server.pluginmanifest.md)

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ohltyler: So `httpClient` and `notifications` are core services, and alerting plugin decided to pass these through `props` since the services didn't have to be passed deep down into the component tree. For AD, we have much more places where these services need to be accessed deeply, and is why I decided to use React context to access them instead. See [here](https://github.com/elastic/kibana/blob/master/src/core/MIGRATION.md#how-can-i-avoid-passing-core-services-deeply-within-my-ui-component-tree) for more info

ohltyler: Regarding the import path, I think there is a way to shorten & customize import paths in the `tsconfig.json` file, although I haven't looked into it much. I think this would be a good thing to look into later, and could change for other common imports as well, rather than having so many of these long relative imports. Would help prevent imports from breaking during refactoring as well.

kibana.json
ohltyler: Yizhe has some changes for this, there will be a follow up PR with upgrade-related changes.

public/pages/DetectorDetail/containers/DetectorDetail.tsx
ftianli-amzn: I think Yaliang agrees with using "context" to access the core services, as in the part of Anomaly Detector feature in Alerting plugin (https://github.com/opendistro-for-elasticsearch/alerting-kibana-plugin/blob/master/public/pages/CreateMonitor/containers/AnomalyDetectors/AnomalyDetectors.js#L24), React context is used instead of "props" 😉. And I think the reason that the import path is not outstanding in alerting plugin is it doesn't needs to declare the type there. 
Anyway, it's good to know that there is a way to shorten the import paths.

ylwu-amzn: Cool, not a big problem, let's keep as is currently.

kibana.json
ylwu-amzn: Sure, make sense.

ylwu-amzn (approved):

Issue #24, opened 2020-04-15
Improve sunburst chart for Anomaly Distribution


yizheliu-amazon:
Given only 1 detector exists, the inside Index circle will have issue in following scenarios:
1. if Linked Label is configured for identifying index, label of Index circle will be invisible;
1. if Fill Label is configured for identifying index, label of it will be overlapped with outer Detector circle

**Screenshot of 1st scenario:**
![Screen Shot 2020-04-15 at 3 28 31 PM](https://user-images.githubusercontent.com/59710443/79395189-fb06bc80-7f2d-11ea-9f0f-5e264937a6d2.png)

**Screenshot of 2nd scenario**:
![Screen Shot 2020-04-15 at 3 25 09 PM](https://user-images.githubusercontent.com/59710443/79394886-57b5a780-7f2d-11ea-903a-5b4bdd854fcb.png)

We think currently it is fine because sunburst chart does not make big difference if there is only 1 detector + 1 index. We will keep digging into this, and may open issue for [elastic-chart](https://github.com/elastic/elastic-charts) to see if we are using it in a correct way or it is bug of the library.

yizheliu-amazon:
we are also seeing the same issue: https://github.com/elastic/elastic-charts/issues/633

yizheliu-amazon:
Opened issue to owner of elastic-charts: https://github.com/elastic/elastic-charts/issues/637

ylwu-amzn:
Test on 7.7, one detector looks like this
![Screen Shot 2020-06-23 at 10 28 03 AM](https://user-images.githubusercontent.com/49084640/85435409-434ce700-b53c-11ea-8c17-25e64a6089f6.png)

yizheliu-amazon:
Closed by this PR: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/260

PR #332, opened 2020-11-24
Fix sunburst chart height is 0

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
In 7.10.0, the Elastic chart library is upgraded to [version 23](https://github.com/elastic/kibana/blob/v7.10.0/package.json#L226), while in 7.9.1, it is still [version 19](https://github.com/elastic/kibana/blob/v7.9.1/package.json#L127). In the new version of Elastic chart library, sunburst chart  has class "echChart" [automatically added to it](https://github.com/elastic/elastic-charts/blob/83919ffe294257839d360b589ce10f405e04af5b/src/components/chart.tsx#L164), while [echChart](https://github.com/elastic/elastic-charts/blob/83919ffe294257839d360b589ce10f405e04af5b/src/components/_container.scss#L1) is using `height 100%`, which causes the height of chart is 0 because its parent component has no height value specified.

Thus the fix is to explicitly add sunburst class as its parent.

Before:
Screen Shot 2020-11-24 at 2 35 34 PM

After:
Screen Shot 2020-11-24 at 2 34 54 PM

### DIFF
--- a/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
+++ b/public/pages/Dashboard/Components/AnomaliesDistribution.tsx
@@ -170,67 +170,65 @@ export const AnomaliesDistributionChart = (
         <EuiFlexGroup justifyContent="center">
           <EuiFlexItem grow={false}>
             {isEmpty(anomalyDistribution) ? null : (
-              <Chart className="anomalies-distribution-sunburst">
-                <Partition
-                  id="Anomalies by index and detector"
-                  data={anomalyDistribution}
-                  valueAccessor={(d: Datum) => d.count as number}
-                  valueFormatter={(d: number) => d.toString()}
-                  layers={[
-                    {
-                      groupByRollup: (d: Datum) => d.indices,
-                      nodeLabel: (d: Datum) => {
-                        return d;
+              <div className="anomalies-distribution-sunburst">
+                <Chart>
+                  <Partition
+                    id="Anomalies by index and detector"
+                    data={anomalyDistribution}
+                    valueAccessor={(d: Datum) => d.count as number}
+                    valueFormatter={(d: number) => d.toString()}
+                    layers={[
+                      {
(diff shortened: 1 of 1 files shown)


public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -170,67 +170,65 @@ export const AnomaliesDistributionChart = (
...
-                        textInvertible: true,
-                      },
-                      shape: {
-                        fillColor: (d) => {
-                          return fillOutColors(
-                            d,
-                            (d.x0 + d.x1) / 2 / (2 * Math.PI),
-                            []
-                          );
+              <div className="anomalies-distribution-sunburst">
```
yizheliu-amazon, line 173: This is the only change to fix the issue: moving `sunburst` class to parent class. No changes to chart related code below

public/pages/Dashboard/Components/AnomaliesDistribution.tsx
```diff
@@ -170,67 +170,65 @@ export const AnomaliesDistributionChart = (
...
-                    fillLabel: {
-                      textInvertible: true,
-                    },
-                    linkLabel: {
-                      maxCount: 0,
-                    },
-                    // TODO: Given only 1 detector exists, the inside Index circle will have issue in following scenarios:
-                    // 1: if Linked Label is configured for identifying index, label of Index circle will be invisible;
-                    // 2: if Fill Label is configured for identifying index, label of it will be overlapped with outer Detector circle
-                    // Issue link: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/24
```
yizheliu-amazon, line 230: This comment is removed because the issue is already fixed by elastic chart library upgrading. Related issue #24 is already closed.

weicongs-amazon (approved):

ohltyler (approved):

PR #335, opened 2020-11-25
change side bar position for anomly detection plugin

zhyuanqi:
*Issue #, if available:*

*Description of changes:*
Change position of Anomaly detection plugin in side bar. 
The order is designed by UX designer and in step of 1000(https://github.com/elastic/kibana/blob/master/docs/development/core/public/kibana-plugin-core-public.appcategory.md). 

image (4)

### DIFF
--- a/public/plugin.ts
+++ b/public/plugin.ts
@@ -19,7 +19,6 @@ import {
   CoreStart,
   Plugin,
   PluginInitializerContext,
-  DEFAULT_APP_CATEGORIES,
 } from '../../../src/core/public';
 import {
   AnomalyDetectionKibanaPluginSetup,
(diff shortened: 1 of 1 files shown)


public/plugin.ts
```diff
@@ -40,8 +40,13 @@ export class AnomalyDetectionKibanaPlugin
     core.application.register({
       id: 'opendistro-anomaly-detection-kibana',
       title: 'Anomaly Detection',
-      category: DEFAULT_APP_CATEGORIES.kibana,
```
yizheliu-amazon, line 43: thanks for the change, can you also remove the unused import `DEFAULT_APP_CATEGORIES` at line 22 after the only usage is removed?

yizheliu-amazon (approved):

public/plugin.ts
zhyuanqi: sure. Thanks for reminding.

ohltyler (approved):
Thanks for the change!

seraphjiang (approved):

PR #339, opened 2020-12-02
Fix 2 issues during recent upgrade

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Fix 2 issues during recent upgrade
- when CategoryField is checked, and then unchecked, in parent <formik> will still have validation error saying category field is empty:
  - root cause is `setFieldValue` will [trigger validation](https://formik.org/docs/api/formik#validateonchange-boolean) if `validationOnChange` to `true`(default value). Changing `validationOnChange` to `false` fix the issue
- backend failure is propagated as success to frontend
 - root cause is returning `kibanaResponse.ok` when error is found like [such line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/routes/ad.ts#L202-L206) will NOT reach to [FAILURE](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/reducers/ad.ts#L148) type in reducer. Instead, it will reach to SUCCESS type. For example, given user has no permission to stop/start detector, after such user starts detector, UX says detector is started, even though backend throws exception, and detector is NOT started.

Before fix
Screen Shot 2020-12-02 at 2 53 03 PM
After fix
Screen Shot 2020-12-02 at 2 55 05 PM

### DIFF
--- a/public/pages/EditFeatures/containers/EditFeatures.tsx
+++ b/public/pages/EditFeatures/containers/EditFeatures.tsx
@@ -225,10 +225,7 @@ export function EditFeatures(props: EditFeaturesProps) {
     } catch (err) {
       core.notifications.toasts.addDanger(
         prettifyErrorMessage(
-          getErrorMessage(
-            get(err, 'body.message'),
-            'There was a problem starting detector job'
-          )
+          getErrorMessage(err, 'There was a problem starting detector job')
         )
       );
     }
@@ -257,10 +254,7 @@ export function EditFeatures(props: EditFeaturesProps) {
     } catch (err) {
       core.notifications.toasts.addDanger(
         prettifyErrorMessage(
-          getErrorMessage(
-            get(err, 'body.message'),
-            'There was a problem updating feature'
-          )
+          getErrorMessage(err, 'There was a problem updating feature')
         )
       );
       setSubmitting(false);
@@ -295,7 +289,7 @@ export function EditFeatures(props: EditFeaturesProps) {
       } else {
         setSaveFeatureOption(SAVE_FEATURE_OPTIONS.START_AD_JOB);
       }
-      if (errors.categoryField) {
+      if (isHCDetector && errors.categoryField) {
         focusOnCategoryField();
         return;
       }
@@ -382,7 +376,6 @@ export function EditFeatures(props: EditFeaturesProps) {
           handleSubmit(values, actions.setSubmitting)
         }
         validate={validateFeatures}
-        validateOnChange={false}
       >
         {({
           values,
--- a/public/pages/EditFeatures/containers/EditFeatures.tsx
+++ b/public/pages/EditFeatures/containers/EditFeatures.tsx
@@ -225,7 +225,10 @@ export function EditFeatures(props: EditFeaturesProps) {
     } catch (err) {
       core.notifications.toasts.addDanger(
         prettifyErrorMessage(
-          getErrorMessage(err, 'There was a problem starting detector job')
+          getErrorMessage(
+            get(err, 'body.message'),
+            'There was a problem starting detector job'
+          )
         )
       );
     }
@@ -254,7 +257,10 @@ export function EditFeatures(props: EditFeaturesProps) {
     } catch (err) {
       core.notifications.toasts.addDanger(
         prettifyErrorMessage(
-          getErrorMessage(err, 'There was a problem updating feature')
+          getErrorMessage(
+            get(err, 'body.message'),
+            'There was a problem updating feature'
+          )
         )
       );
       setSubmitting(false);
@@ -376,6 +382,7 @@ export function EditFeatures(props: EditFeaturesProps) {
           handleSubmit(values, actions.setSubmitting)
         }
         validate={validateFeatures}
+        validateOnChange={false}
       >
         {({
           values,
--- a/public/redux/middleware/__tests__/clientMiddleware.test.ts
+++ b/public/redux/middleware/__tests__/clientMiddleware.test.ts
@@ -92,12 +92,12 @@ describe('client MiddleWare', () => {
     const { next, invoke } = create();
     const actionCall = jest
       .fn()
-      .mockReturnValue(fooPromise(true, { ok: false }));
+      .mockReturnValue(fooPromise(true, { body: { ok: false } }));
     const action = { type: 'TEST', request: actionCall };
     return invoke(action).then(() => {
       expect(next.mock.calls.length).toBe(2);
       expect(next.mock.calls[1][0]).toEqual({
-        result: { ok: false },
+        result: { body: { ok: false } },
         type: 'TEST_SUCCESS',
       });
     });
--- a/server/routes/utils/adHelpers.ts
+++ b/server/routes/utils/adHelpers.ts
@@ -19,6 +19,7 @@ import { GetDetectorsQueryParams } from '../../models/types';
 import { mapKeysDeep, toCamel, toSnake } from '../../utils/helpers';
 import { DETECTOR_STATE } from '../../utils/constants';
 import { InitProgress } from '../../models/interfaces';
+import { kibanaResponseFactory } from '../../../../../src/core/server/http/router';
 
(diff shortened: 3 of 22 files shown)


public/pages/EditFeatures/containers/EditFeatures.tsx
```diff
@@ -376,6 +382,7 @@ export function EditFeatures(props: EditFeaturesProps) {
           handleSubmit(values, actions.setSubmitting)
         }
         validate={validateFeatures}
+        validateOnChange={false}
```
yizheliu-amazon, line 385: This is the only change to fix formik validation issue for CategoryField

yizheliu-amazon:
Have verified unit test and e2e test passed.

yizheliu-amazon:
will add change for release note in next commit.

ohltyler:
Regarding the issue of returning `kibanaResponse.ok` in the failure case, did you ever look into changing [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L44) and [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L49)?

Looks like I may have forgotten to remove the `data.` prefix there, which was changed with the new HTTP client. For example, I've tested locally and gotten it to properly go to the `FAILURE` reducer type by changing `data.ok` to `ok`

public/pages/EditFeatures/containers/EditFeatures.tsx
ylwu-amzn: Does this `validateOnChange` only control validating CategoryField or not? Just make sure we don't ignore validation for other parts which necessary.

server/routes/utils/adHelpers.ts
```diff
@@ -253,3 +254,55 @@ export const getErrorMessage = (err: any) => {
...
     : get(err, 'message');
 };
+
+export const getKibanaErrorResponseByStatusCode = (
+  statusCode: number,
+  message: string
+) => {
+  switch (statusCode) {
+    case 400:
+      return kibanaResponseFactory.badRequest({
```
ylwu-amzn, line 264: Just curious, why need to explicitly return `kibanaResponseFactory.badRequest`, looks like something duplicate for `400` and `badRequest` Is there any general error response to accept status code ?

public/pages/EditFeatures/containers/EditFeatures.tsx
yizheliu-amazon: Yes. For `Features`, its `validateOnChange` is set to be true in [code here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/pages/EditFeatures/containers/EditFeatures.tsx#L155). And when save button is clicked, it will do check for category field again in [handleSaveChanges](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/pages/EditFeatures/containers/EditFeatures.tsx#L278).

yizheliu-amazon:
> Regarding the issue of returning `kibanaResponse.ok` in the failure case, did you ever look into changing [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L44) and [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L49)?
> 
> Looks like I may have forgotten to remove the `data.` prefix there, which was changed with the new HTTP client. For example, I've tested locally and gotten it to properly go to the `FAILURE` reducer type by changing `data.ok` to `ok`, since it is now finding the `ok: false` field inside of the Kibana response.
> 
> I believe the same goes for changing `data.error` to `error`.

Good catch. Removing `data.` prefix works.

server/routes/utils/adHelpers.ts
yizheliu-amazon: `kibanaResponseFactory` has multiple response items with corresponding statusCode, `badRequest` has 400 status code. as shown like [this](https://github.com/elastic/kibana/blob/v7.10.0/src/core/server/http/router/response.ts#L155). This method is for given status code, return corresponding response.

But given Tyler's found above, I will revert this change. No need to worry about it.

ohltyler:
> > Regarding the issue of returning `kibanaResponse.ok` in the failure case, did you ever look into changing [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L44) and [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L49)?
> > Looks like I may have forgotten to remove the `data.` prefix there, which was changed with the new HTTP client. For example, I've tested locally and gotten it to properly go to the `FAILURE` reducer type by changing `data.ok` to `ok`, since it is now finding the `ok: false` field inside of the Kibana response.
> > I believe the same goes for changing `data.error` to `error`.
> 
> Good catch. Removing `data.` prefix works.

Cool, so seems we can revert the Kibana response changes on server side, and the error parsing changes on the browser side then?

yizheliu-amazon:
> > > Regarding the issue of returning `kibanaResponse.ok` in the failure case, did you ever look into changing [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L44) and [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/redux/middleware/clientMiddleware.ts#L49)?
> > > Looks like I may have forgotten to remove the `data.` prefix there, which was changed with the new HTTP client. For example, I've tested locally and gotten it to properly go to the `FAILURE` reducer type by changing `data.ok` to `ok`, since it is now finding the `ok: false` field inside of the Kibana response.
> > > I believe the same goes for changing `data.error` to `error`.
> > 
> > 
> > Good catch. Removing `data.` prefix works.
> 
> Cool, so seems we can revert the Kibana response changes on server side, and the error parsing changes on the browser side then? Apologize for missing this on the initial commits.

Right, I will revert those lines and make it way simpler.

No worries. I think it is not easy for find such issue, another data point for more automated test.

public/pages/EditFeatures/containers/EditFeatures.tsx
yizheliu-amazon: sorry for reverting the change. has found it can not fix the issue completely. Has made such change to completely fix the issue: https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/339/files#diff-b12f07c85b6d89dbdd055f0482e17c93168efdabb5b2d08ec6fde29ee7b6dfddR292

public/redux/middleware/__tests__/clientMiddleware.test.ts
```diff
@@ -92,12 +92,12 @@ describe('client MiddleWare', () => {
...
     const actionCall = jest
       .fn()
-      .mockReturnValue(fooPromise(true, { ok: false }));
+      .mockReturnValue(fooPromise(true, { body: { ok: false } }));
     const action = { type: 'TEST', request: actionCall };
     return invoke(action).then(() => {
       expect(next.mock.calls.length).toBe(2);
       expect(next.mock.calls[1][0]).toEqual({
-        result: { ok: false },
+        result: { body: { ok: false } },
```
ohltyler, line 100: Are these changes still necessary?

yizheliu-amazon: yes. This is to fix the failed ut.

ohltyler (approved):
Thanks for the fix!

ylwu-amzn (approved):

PR #340, opened 2020-12-03
Use plotly.js-dist to reduce artifact size

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Use plotly.js-dist library to reduce artifact size. Similar as https://github.com/opendistro-for-elasticsearch/kibana-visualizations/pull/5

Size of build artifact is reduced from 50M to 11M.

UT and e2e test passed. Also verified heatmap works for HC detector by installing artifact to binary Kibana. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "brace": "0.11.1",
     "formik": "^2.2.5",
     "lodash": "^4.17.20",
-    "plotly.js": "^1.55.2",
+    "plotly.js-dist": "^1.57.1",
     "react-plotly.js": "^2.4.0",
     "react-redux": "^7.2.0",
     "reselect": "^4.0.0"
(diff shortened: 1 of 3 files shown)


weicongs-amazon (approved):
just curious. Why does this dist one have smaller size?

ylwu-amzn:
> just curious. Why does this dist one have smaller size?

Generally dist one will do some compression like remove extra space, line-breaks. I think it's the same for plotly.

ylwu-amzn (approved):

yizheliu-amazon:
> just curious. Why does this dist one have smaller size?

Didn;t find good info on such question via google. But from the npm website, we can see the difference: Unpacked size is over 60M for [plotly.js](https://www.npmjs.com/package/plotly.js), while size is around 7M for [plotly.js-dist](https://www.npmjs.com/package/plotly.js-dist). It looks to me `plotly.js-dist` has only necessary modules, while `plotly.js` has everything.

ohltyler (approved):

Issue #313, opened 2020-10-16
Improve AnomalyHistory Chart performance for HC detector


yizheliu-amazon:
Currently for long date range result in AnomalyHistoryChart, we use bucketized query, which split time range into up to 1000 windows, and only takes max anomaly grade in each small time window(across multiple intervals). Query is like [this](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/utils/anomalyResultUtils.ts#L349-L399). The benefit of it is that we can have up to only 1000 points even for long date range, and it can have good performance for rendering; downside of it is it can miss anomalies if in the small time window there are more than 1 anomaly results. 

For HC detector heatmap, we need to show accurate occurrence/count of anomalies for each cell, but such bucketized query can give us less anomalies than actual number. 

Current implementation is to get raw anomaly results, i.e. all results in the date range, but this can potentially bring in performance issue. Need to have better implementation for this.

ohltyler:
@yizheliu-amazon based on #350 can this be closed?

yizheliu-amazon:
Thanks for catching this. Yes. It can be closed.

PR #350, opened 2020-12-18
Refactor AnomalyHistory Chart to improve performance for HC detector

yizheliu-amazon:
*Issue #, if available:*
#313 
*Description of changes:*
Currently, AnomalyHistory chart loads all anomaly only data and pass it to heatmap chart to do the filtering/sorting for rendering anomaly summaries data on heatmap. The downside of it is 1) we have to manually aggregate/sort raw anomaly data, and it can cause performance issue if there are too many anomalies in specified date range; 2) we miss the feature data of anomaly grade 0 data points, such that it has less friendly UX compared to all feature data is rendered

This PR changes to 1) find anomalous entities via ES query; 2) get anomaly summaries for each entities via ES query;3) get anomaly/feature data only when heatmap cell is selected. 

This moves the heavy aggregation/sorting work to ES from front-end, and has better UX with all feature data is rendered. 

After the change:
Preview: same as before
Preview_selected

DetectorResult page: anomaly data points with anomaly grade 0 is also included
Result_selected_anomaly

DetectorResult page: feature data points with anomaly grade 0 is also included
Result_selected_feature

UT will be added in following commit.

### DIFF
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -24,6 +24,7 @@ import {
 import { get } from 'lodash';
 import moment, { DurationInputArg2 } from 'moment';
 import React, { useState } from 'react';
+import { EntityAnomalySummaries } from '../../../../server/models/interfaces';
 import ContentPanel from '../../../components/ContentPanel/ContentPanel';
 import { useDelayedLoader } from '../../../hooks/useDelayedLoader';
 import {
@@ -38,6 +39,7 @@ import { AnomalyDetailsChart } from '../containers/AnomalyDetailsChart';
 import {
   AnomalyHeatmapChart,
   HeatmapCell,
+  HeatmapDisplayOption,
 } from '../containers/AnomalyHeatmapChart';
 import {
   getAnomalyGradeWording,
@@ -71,10 +73,13 @@ interface AnomaliesChartProps {
   isHCDetector?: boolean;
   detectorCategoryField?: string[];
   onHeatmapCellSelected?(heatmapCell: HeatmapCell): void;
+  onDisplayOptionChanged?(heatmapDisplayOption: HeatmapDisplayOption): void;
   selectedHeatmapCell?: HeatmapCell;
   newDetector?: Detector;
   zoomRange?: DateRange;
   anomaliesResult: Anomalies | undefined;
+  heatmapDisplayOption?: HeatmapDisplayOption;
+  entityAnomalySummaries?: EntityAnomalySummaries[];
 }
 
 export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
@@ -183,7 +188,9 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
         <EuiFlexGroup direction="column">
           {props.isHCDetector &&
           props.onHeatmapCellSelected &&
-          props.detectorCategoryField ? (
+          props.detectorCategoryField &&
+          props.onDisplayOptionChanged &&
+          props.entityAnomalySummaries ? (
             <EuiFlexGroup style={{ padding: '20px' }}>
               <EuiFlexItem style={{ margin: '0px' }}>
                 <div
@@ -222,6 +229,10 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                           'detectionInterval.period.unit'
                         )}
                         onHeatmapCellSelected={props.onHeatmapCellSelected}
+                        entityAnomalySummaries={props.entityAnomalySummaries}
+                        onDisplayOptionChanged={props.onDisplayOptionChanged}
+                        //@ts-ignore
+                        heatmapDisplayOption={props.heatmapDisplayOption}
                       />,
                       props.showAlerts !== true
                         ? [
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -189,8 +189,8 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
           {props.isHCDetector &&
           props.onHeatmapCellSelected &&
           props.detectorCategoryField &&
-          props.onDisplayOptionChanged &&
-          props.entityAnomalySummaries ? (
+          (props.showAlerts !== true ||
+            (props.onDisplayOptionChanged && props.entityAnomalySummaries)) ? (
             <EuiFlexGroup style={{ padding: '20px' }}>
               <EuiFlexItem style={{ margin: '0px' }}>
                 <div
@@ -231,7 +231,6 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                         onHeatmapCellSelected={props.onHeatmapCellSelected}
                         entityAnomalySummaries={props.entityAnomalySummaries}
                         onDisplayOptionChanged={props.onDisplayOptionChanged}
-                        //@ts-ignore
                         heatmapDisplayOption={props.heatmapDisplayOption}
                       />,
                       props.showAlerts !== true
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -177,6 +177,17 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
     );
   };
 
+  const hasValidHCProps = () => {
+    return (
+      props.isHCDetector &&
+      props.onHeatmapCellSelected &&
+      props.detectorCategoryField &&
+      (props.showAlerts !== true ||
+        (props.showAlerts &&
+          props.onDisplayOptionChanged &&
+          props.entityAnomalySummaries))
+    );
+  };
   return (
     <React.Fragment>
       <ContentPanel
@@ -186,13 +197,7 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
         }
       >
         <EuiFlexGroup direction="column">
-          {props.isHCDetector &&
-          props.onHeatmapCellSelected &&
-          props.detectorCategoryField &&
-          (props.showAlerts !== true ||
-            (props.showAlerts &&
-              props.onDisplayOptionChanged &&
-              props.entityAnomalySummaries)) ? (
+          {hasValidHCProps() ? (
             <EuiFlexGroup style={{ padding: '20px' }}>
               <EuiFlexItem style={{ margin: '0px' }}>
                 <div
@@ -230,10 +235,14 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
                           props.detector,
                           'detectionInterval.period.unit'
                         )}
+                        //@ts-ignore
                         onHeatmapCellSelected={props.onHeatmapCellSelected}
                         entityAnomalySummaries={props.entityAnomalySummaries}
                         onDisplayOptionChanged={props.onDisplayOptionChanged}
                         heatmapDisplayOption={props.heatmapDisplayOption}
+                        // TODO use props.isNotSample after Tyler's change is merged
+                        // https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r547009140
+                        isNotSample={props.showAlerts === true}
                       />,
                       props.showAlerts !== true
                         ? [
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -190,7 +190,9 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
           props.onHeatmapCellSelected &&
           props.detectorCategoryField &&
           (props.showAlerts !== true ||
-            (props.onDisplayOptionChanged && props.entityAnomalySummaries)) ? (
+            (props.showAlerts &&
+              props.onDisplayOptionChanged &&
+              props.entityAnomalySummaries)) ? (
             <EuiFlexGroup style={{ padding: '20px' }}>
               <EuiFlexItem style={{ margin: '0px' }}>
                 <div
--- a/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
@@ -182,6 +182,10 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
       props.isHCDetector &&
       props.onHeatmapCellSelected &&
       props.detectorCategoryField &&
+      // For Non-Sample HC detector case, aka realtime HC detector(showAlert == true),
+      // we use anomaly summaries data to render heatmap
+      // we must have function onDisplayOptionChanged and entityAnomalySummaries defined
+      // so that heatmap can work as expected.
       (props.showAlerts !== true ||
         (props.showAlerts &&
           props.onDisplayOptionChanged &&
--- a/public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
+++ b/public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
@@ -58,9 +58,9 @@ interface AnomalyHeatmapChartProps {
   detectorInterval?: number;
   unit?: string;
   onHeatmapCellSelected(cell: HeatmapCell | undefined): void;
-  onDisplayOptionChanged(option: HeatmapDisplayOption | undefined): void;
-  heatmapDisplayOption: HeatmapDisplayOption;
-  entityAnomalySummaries: EntityAnomalySummaries[];
+  onDisplayOptionChanged?(option: HeatmapDisplayOption | undefined): void;
+  heatmapDisplayOption?: HeatmapDisplayOption;
+  entityAnomalySummaries?: EntityAnomalySummaries[];
 }
 
 export interface HeatmapCell {
@@ -80,13 +80,6 @@ const COMBINED_OPTIONS = {
     { label: 'Top 20', value: 20 },
     { label: 'Top 30', value: 30 },
   ],
-};
-
-export const isCombinedViewEntityOption = (inputOption: any) => {
-  const combinedOptionsLabels = COMBINED_OPTIONS.options.map((option) =>
-    get(option, 'label', '')
-  );
-  return combinedOptionsLabels.includes(get(inputOption, 'label', ''));
 };
 
 export const INITIAL_HEATMAP_DISPLAY_OPTION = {
@@ -139,25 +132,22 @@ export const AnomalyHeatmapChart = React.memo(
       ];
     };
 
-    // const [originalHeatmapData, setOriginalHeatmapData] = useState(
-    //   getAnomaliesHeatmapData(
-    //     props.anomalies,
-    //     props.dateRange,
-    //     AnomalyHeatmapSortType.SEVERITY,
-    //     COMBINED_OPTIONS.options[0].value
-    //   )
-    // );
-
     const [originalHeatmapData, setOriginalHeatmapData] = useState(
-      getEnitytAnomaliesHeatmapData(
-        props.dateRange,
-        props.entityAnomalySummaries,
-        props.heatmapDisplayOption.entityOption.value
-      )
+      props.showAlerts
+        ? // use anomaly summary data in case of realtime result
+          getEnitytAnomaliesHeatmapData(
+            props.dateRange,
+            props.entityAnomalySummaries,
+            props.heatmapDisplayOption.entityOption.value
+          )
+        : // use anomalies data in case of sample result
+          getAnomaliesHeatmapData(
+            props.anomalies,
+            props.dateRange,
+            AnomalyHeatmapSortType.SEVERITY,
+            COMBINED_OPTIONS.options[0].value
+          )
     );
-
-    // console.log('originalHeatmapData', originalHeatmapData);
-    // console.log('originalHeatmapDataCopy', originalHeatmapDataCopy);
 
     const [heatmapData, setHeatmapData] = useState<PlotData[]>(
       originalHeatmapData
@@ -165,7 +155,11 @@ export const AnomalyHeatmapChart = React.memo(
 
     const [sortByFieldValue, setSortByFieldValue] = useState<
       AnomalyHeatmapSortType
-    >(props.heatmapDisplayOption.sortType);
+    >(
+      props.showAlerts
+        ? props.heatmapDisplayOption.sortType
+        : SORT_BY_FIELD_OPTIONS[0].value
+    );
 
     const [currentViewOptions, setCurrentViewOptions] = useState([
       getViewEntityOptions(originalHeatmapData)[0].options[0],
@@ -262,22 +256,25 @@ export const AnomalyHeatmapChart = React.memo(
       if (isEmpty(selectedViewOptions)) {
         // when `clear` is hit for combo box
         setCurrentViewOptions([COMBINED_OPTIONS.options[0]]);
-        const displayTopEntityNum = get(COMBINED_OPTIONS.options[0], 'value');
 
-        props.onDisplayOptionChanged({
-          sortType: sortByFieldValue,
-          entityOption: COMBINED_OPTIONS.options[0],
-        });
-        // const updateHeatmapPlotData = getAnomaliesHeatmapData(
-        //   props.anomalies,
-        //   props.dateRange,
-        //   sortByFieldValue,
-        //   displayTopEntityNum
-        // );
-        // setOriginalHeatmapData(updateHeatmapPlotData);
-        // setHeatmapData(updateHeatmapPlotData);
-        // setNumEntities(updateHeatmapPlotData[0].y.length);
-        // setEntityViewOptions(getViewEntityOptions(updateHeatmapPlotData));
+        if (props.showAlerts) {
+          props.onDisplayOptionChanged({
+            sortType: sortByFieldValue,
+            entityOption: COMBINED_OPTIONS.options[0],
+          });
+        } else {
+          const displayTopEntityNum = get(COMBINED_OPTIONS.options[0], 'value');
+          const updateHeatmapPlotData = getAnomaliesHeatmapData(
+            props.anomalies,
+            props.dateRange,
+            sortByFieldValue,
+            displayTopEntityNum
+          );
+          setOriginalHeatmapData(updateHeatmapPlotData);
+          setHeatmapData(updateHeatmapPlotData);
+          setNumEntities(updateHeatmapPlotData[0].y.length);
+          setEntityViewOptions(getViewEntityOptions(updateHeatmapPlotData));
+        }
         return;
       }
       const nonCombinedOptions = [] as any[];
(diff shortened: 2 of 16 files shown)


public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -183,7 +188,11 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
         <EuiFlexGroup direction="column">
           {props.isHCDetector &&
           props.onHeatmapCellSelected &&
-          props.detectorCategoryField ? (
+          props.detectorCategoryField &&
+          (props.showAlerts !== true ||
+            (props.showAlerts &&
+              props.onDisplayOptionChanged &&
+              props.entityAnomalySummaries)) ? (
```
ohltyler, line 195: Could you abstract this out to some boolean var for readability and to understand what this means exactly?

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -41,8 +41,10 @@ import {
   AnomalyHeatmapSortType,
   sortHeatmapPlotData,
   filterHeatmapPlotDataByY,
+  getEnitytAnomaliesHeatmapData,
```
ohltyler, line 44: Minor: typo in fn name here

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -109,28 +124,57 @@ export const AnomalyHeatmapChart = React.memo(
...
+        ),
         {
           label: 'Individual entities',
-          options: individualEntityOptions,
+          options: individualEntityOptions.reverse(),
         },
       ];
     };
 
+    const getCombindeOptions = (
```
ohltyler, line 138: Minor: typo in fn name

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -109,28 +124,57 @@ export const AnomalyHeatmapChart = React.memo(
...
+    };
+
     const [originalHeatmapData, setOriginalHeatmapData] = useState(
-      getAnomaliesHeatmapData(
-        props.anomalies,
-        props.dateRange,
-        AnomalyHeatmapSortType.SEVERITY,
-        COMBINED_OPTIONS.options[0].value
-      )
+      props.showAlerts
```
ohltyler, line 152: Can we add another prop to determine if we are using actual vs. sample results, rather than reusing the `showAlerts` one? One reason for this is that those may not always coincide with each other. For example, with historical detector results, `props.showAlerts` is `false` since alerts don't apply, but it is also not sample results.

I'm reusing the `AnomalyHistory` component, and have actually added a prop to handle this (as well as for all of the necessary children components) [here](https://github.com/ohltyler/anomaly-detection-kibana-plugin/commit/33b4c0c73364e6ed86dcfa4102032134d6d7fb49). I'm wondering if you can maybe do something similar here. Will help with possible merge conflicts down the road as well. Let me know what you think.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -193,14 +237,14 @@ export const AnomalyHeatmapChart = React.memo(
...
+          const selectedStartDate = moment(
             //@ts-ignore
             heatmapData[0].x[selectedCellIndices[1]],
             HEATMAP_X_AXIS_DATE_FORMAT
           ).valueOf();
 
-          const selectedStartDate =
-            selectedEndDate -
+          const selectedEndDate =
+            selectedStartDate +
```
ohltyler, line 247: Could you explain why the start and end dates are switched here?

yizheliu-amazon: This is because I make [change](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350/files#diff-cde386f024d306c06003870e2bde6d912b144586cbd972d40e77a00defddb529R273) switch the date of each cell in heatmap from endDate to startDate. I think showing the start date of heatmap cell as `Time` in the hover box is better.

Thus, the `x` of heatmap data here is changed to startDate, and I have to switch endDate and startDate here

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
yizheliu-amazon: sure. will do that.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: thanks for the catch. Will change it.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: Yeah, I was thinking about the same thing. Sure. I can add another prop for this case, something like `isUsingSampleAnomaly`

server/routes/ad.ts
```diff
@@ -715,6 +719,34 @@ export default class AdService {
...
                 },
               },
+              ...(entityName && entityValue
+                ? [
+                    {
+                      nested: {
+                        path: 'entity',
+                        query: {
+                          term: {
+                            'entity.name': {
```
ylwu-amzn, line 729: User [`ENTITY_NAME_PATH_FIELD`](ENTITY_NAME_PATH_FIELD) ?
Same question for `'entity.value'`

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -858,3 +937,245 @@ export const filterWithHeatmapFilter = (
...
+      [TOP_ENTITIES_FIELD]: {
+        nested: {
+          path: ENTITY_FIELD,
+        },
+        aggs: {
+          [TOP_ENTITY_AGGS]: {
+            terms: {
+              field: ENTITY_VALUE_PATH_FIELD,
+              size: size,
+              ...(sortType === AnomalyHeatmapSortType.SEVERITY
```
ylwu-amzn, line 986: What if user use `AnomalyHeatmapSortType.OCCURRENCES`  ?

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -295,6 +300,114 @@ export const getAnomaliesHeatmapData = (
...
+      zmin: 0,
+      zmax: 1,
+      type: 'heatmap',
+      showscale: false,
+      xgap: 2,
+      ygap: 2,
+      opacity: 1,
+      text: numAnomalyGrades.reverse(),
+      hovertemplate:
+        '<b>Time</b>: %{x}<br>' +
```
ylwu-amzn, line 402: Can we abstract these lines into const or function?

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: The default order is sorting by `doc_count` descending, aka by OCCURRENCES, as per [Elastic doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-order). Thus, there is no need to specify query for `OCCURRENCES` case

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
yizheliu-amazon: sure. will do that.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -61,6 +61,7 @@ interface AnomalyHeatmapChartProps {
   onDisplayOptionChanged?(option: HeatmapDisplayOption | undefined): void;
   heatmapDisplayOption?: HeatmapDisplayOption;
   entityAnomalySummaries?: EntityAnomalySummaries[];
+  isNotSample: boolean;
```
ohltyler, line 64: You may make this optional for any backwards compatibility issues.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
ohltyler: Got it, sounds good.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
yizheliu-amazon: sure.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -172,6 +177,17 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
     );
   };
 
+  const hasValidHCProps = () => {
+    return (
+      props.isHCDetector &&
```
ylwu-amzn, line 182: Why we need to check so many props ? Can we use less props to check?

yizheliu-amazon: It is to make sure user of `AnomaliesChart` doesn't miss needed props for HC detector, these props are optional because for non-HC detector, such props are not used. I can add comment for the condition around line 185 which can be not straightforward for understanding. Feel free to let me know your thoughts.

ylwu-amzn: Got it, please add some more comments for line 185-188

ylwu-amzn (approved):

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -356,7 +406,8 @@ export const getBucketizedAnomalyResultsQuery = (
   startTime: number,
   endTime: number,
   interval: number,
-  detectorId: string
+  detectorId: string,
+  entity: Entity | undefined = undefined
 ) => {
   const fixedInterval = Math.ceil(
     (endTime - startTime) / (interval * MIN_IN_MILLI_SECS * MAX_DATA_POINTS)
```
kaituo, line 413: Could you help me understand the formula?  I thought we want to find out if the interval is smaller than MAX_DATA_POINTS or not? So

```
let fixedInterval = Math.ceil((endTime - startTime) / (interval * MIN_IN_MILLI_SECS));
if (fixedInterval > MAX_DATA_POINTS) {
    fixedInterval = MAX_DATA_POINTS;
}
```

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -99,58 +118,70 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
...
+        `Failed to get anomaly results for ${props.detector.id}`,
+        err
+      );
+    } finally {
+      setIsLoadingAnomalyResults(false);
     }
+  }
 
+  useEffect(() => {
     fetchRawAnomalyResults(isHCDetector);
```
kaituo, line 180: From line 208, the parameter's name is actually showLoading.  Why is isHCDetector related to showLoading?

kaituo (commented):
I finished half of the code with a lot of basic questions.  Please help me.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -172,6 +177,17 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
...
+      props.onHeatmapCellSelected &&
+      props.detectorCategoryField &&
+      (props.showAlerts !== true ||
+        (props.showAlerts &&
+          props.onDisplayOptionChanged &&
+          props.entityAnomalySummaries))
+    );
```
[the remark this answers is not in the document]
yizheliu-amazon: Yeas, your understanding is right. As per above [comment](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r547009140), I will change from `props.showAlerts` to `props.isNotSample` after Tyler's change is merged. 

More context is: for now `props.showAlerts` is a flag mainly for indicating whether the `chart` is used for Preview result case, aka Sample result case. If false, it is for Sample: if yes, it is for Real detection result. It has such name because we only show alert button on real detection result page, not on Sample result one. But this is not appropriate name because such flag is being used for 2 purposes(determining if chart is for Sample, and if chart should show alert button). That's why Tyler has [change](https://github.com/ohltyler/anomaly-detection-kibana-plugin/commit/33b4c0c73364e6ed86dcfa4102032134d6d7fb49) to refactor use of `showAlerts`, and I will use new `props. isNotSample` once his change is checked in.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -41,28 +41,53 @@ import {
...
 
 interface AnomalyHeatmapChartProps {
   title: string;
   detectorId: string;
   detectorName: string;
-  anomalies: any[];
+  anomalies?: any[];
   dateRange: DateRange;
   isLoading: boolean;
   showAlerts?: boolean;
```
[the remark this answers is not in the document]
yizheliu-amazon: good catch. it was used in previous commit before changing to `isNotSample`. Will remove it.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -109,28 +125,57 @@ export const AnomalyHeatmapChart = React.memo(
...
       return [
-        COMBINED_OPTIONS,
+        getViewableCombinedOptions(
+          COMBINED_OPTIONS,
+          props.heatmapDisplayOption?.entityOption
+        ),
         {
           label: 'Individual entities',
-          options: individualEntityOptions,
+          options: individualEntityOptions.reverse(),
```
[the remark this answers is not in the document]
yizheliu-amazon: Input of the method `inputHeatmapData` has entities in ascending order by sorting field. And then the `individualEntityOptions` array is in ascending order, but we want to list those options in combo box in descending because heatmap is showing data in descending order. 

The reason why we make `inputHeatmapData` in ascending order, to make sure heatmap shows data in descending order, is because in plotly library, the origin of chart is at left bottom. If we make `heatmapData` in descending order, the biggest item will be shown on bottom, while smallest on top, but this looks ascending.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -109,28 +125,57 @@ export const AnomalyHeatmapChart = React.memo(
...
+      props.isNotSample
+        ? // use anomaly summary data in case of realtime result
+          getEntitytAnomaliesHeatmapData(
+            props.dateRange,
+            props.entityAnomalySummaries,
+            props.heatmapDisplayOption.entityOption.value
+          )
+        : // use anomalies data in case of sample result
+          getAnomaliesHeatmapData(
+            props.anomalies,
```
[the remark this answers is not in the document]
yizheliu-amazon: I am not worried about checking because `anomalies` is set at only 1 [place](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/AnomalyCharts/containers/AnomaliesChart.tsx#L86), and it won't be undefined.

After second thought, I think I can make change to check if it exists or not when converting it to heatmap data during `getAnomaliesHeatmapData`, and `props.anomalies` is being only used in `getAnomaliesHeatmapData`

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -265,36 +271,148 @@ export const getAnomaliesHeatmapData = (
...
+
+  entitiesAnomalySummaries.forEach((entityAnomalySummaries) => {
+    const maxAnomalyGradesForEntity = [] as number[];
+    const numAnomalyGradesForEntity = [] as number[];
+
+    const entityValue = get(
+      entityAnomalySummaries,
+      ENTITY_VALUE_PATH_FIELD,
+      ''
+    );
```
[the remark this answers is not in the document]
yizheliu-amazon: I can add it.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
```diff
@@ -221,7 +239,14 @@ export const AnomaliesChart = React.memo((props: AnomaliesChartProps) => {
...
                           'detectionInterval.period.unit'
                         )}
+                        //@ts-ignore
                         onHeatmapCellSelected={props.onHeatmapCellSelected}
+                        entityAnomalySummaries={props.entityAnomalySummaries}
+                        onDisplayOptionChanged={props.onDisplayOptionChanged}
+                        heatmapDisplayOption={props.heatmapDisplayOption}
+                        // TODO use props.isNotSample after Tyler's change is merged
+                        // https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r547009140
+                        isNotSample={props.showAlerts === true}
```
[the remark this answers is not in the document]
yizheliu-amazon: I think above [comment](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r550298360) can explain.

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -265,36 +271,148 @@ export const getAnomaliesHeatmapData = (
...
+
+  const plotTimes = timeWindows.map((timeWindow) => timeWindow.startDate);
+  const timeStamps = plotTimes.map((timestamp) =>
+    moment(timestamp).format(HEATMAP_X_AXIS_DATE_FORMAT)
+  );
+  const plotData = buildHeatmapPlotData(
+    timeStamps,
+    entityValues.reverse(),
+    maxAnomalyGrades.reverse(),
+    numAnomalyGrades.reverse(),
```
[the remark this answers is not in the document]
yizheliu-amazon: the original order is descending order, after `reverse` it is ascending. The reason is explained [above](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r550302879)

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -265,36 +271,148 @@ export const getAnomaliesHeatmapData = (
...
+      // only 1 whitesapce for all entities, to avoid heatmap with single row
+      const blankStrValue = buildBlankStringWithLength(i);
+      entitiesAnomalySummaries.push({
+        entity: {
+          value: blankStrValue,
+        },
+      } as EntityAnomalySummaries);
+    }
+  } else {
+    entitiesAnomalySummaries = cloneDeep(entitiesAnomalySummaryResult);
```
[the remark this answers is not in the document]
yizheliu-amazon: after double confirm, you are right, deep clone can be removed. will remove it.

public/pages/AnomalyCharts/containers/AnomalyHeatmapChart.tsx
```diff
@@ -109,28 +125,57 @@ export const AnomalyHeatmapChart = React.memo(
...
+    const getViewableCombinedOptions = (
+      existingOptions: any,
+      selectedCombinedOption: any | undefined
+    ) => {
+      if (!selectedCombinedOption) {
+        return existingOptions;
+      }
+      return {
+        label: existingOptions.label,
+        options: uniq([selectedCombinedOption, ...existingOptions.options]),
```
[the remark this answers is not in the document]
yizheliu-amazon: Yes. The reason is I want to put the selected combined option as the 1st element of options, because we by default use 1st element as selected option as shown in [code](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350/files#diff-a044985587be3c93885ad08b0bb90396e88600895b94ace004c30cda74f955d7R182)

public/pages/AnomalyCharts/utils/anomalyChartUtils.ts
```diff
@@ -265,36 +271,148 @@ export const getAnomaliesHeatmapData = (
...
+  );
   const resultPlotData = sortHeatmapPlotData(plotData, sortType, displayTopNum);
   return [resultPlotData];
 };
 
+const buildHeatmapPlotData = (
+  x: any[],
+  y: any[],
+  z: any[],
+  text: any[],
```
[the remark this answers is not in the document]
yizheliu-amazon: I would prefer keeping z/text because that method is for building heatmap plot data, it should be coupled with `PlotData` attribute. Next time if we has another use case using heatmap chart, but not with maxAnomalyGrades/numAnomalyGrades, signature of this method can still be appropriate.

server/models/interfaces.ts
```diff
@@ -1,3 +1,22 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
```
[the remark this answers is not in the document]
yizheliu-amazon: Good catch. Will change it.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -77,19 +95,22 @@ export const getLiveAnomalyResults = (
...
+  anomalyOnly: boolean = false,
+  entity: Entity | undefined = undefined
 ) => {
   return {
     from: 0,
     size: MAX_ANOMALIES,
     sortDirection: SORT_DIRECTION.DESC,
     sortField: AD_DOC_FIELDS.DATA_START_TIME,
-    dateRangeFilter: {
+    dateRangeFilter: JSON.stringify({
```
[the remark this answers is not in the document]
yizheliu-amazon: This can be parsed and converted in [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/routes/ad.ts#L725-L727). Actually this change fixes bug which is not found before: dateRangeFilter is not parsed and passed to ES.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -858,3 +937,245 @@ export const filterWithHeatmapFilter = (
...
+              [TOP_ANOMALY_GRADE_SORT_AGGS]: {
+                reverse_nested: {},
+                aggs: {
+                  [MAX_ANOMALY_AGGS]: {
+                    max: {
+                      field: AD_DOC_FIELDS.ANOMALY_GRADE,
+                    },
+                  },
+                },
+              },
```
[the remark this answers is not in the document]
yizheliu-amazon: Yes. We need such aggregation to return `max_anomaly` value. Both OCCURRENCES/SEVERITY case need `anomaly_count` and `max_anomaly` value for each bucket. Only difference is OCCURRENCES sorts by `anomaly_count`, and SEVERITY sorts by `max_anomaly`

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -858,3 +937,245 @@ export const filterWithHeatmapFilter = (
...
+): EntityAnomalySummaries[] => {
+  const rawEntityAnomalySummaries = get(
+    result,
+    `response.aggregations.${TOP_ENTITIES_FIELD}.${TOP_ENTITY_AGGS}.buckets`,
+    []
+  ) as any[];
+  let topEntityAnomalySummaries = [] as EntityAnomalySummaries[];
+  rawEntityAnomalySummaries.forEach((item) => {
+    const anomalyCount = get(item, 'doc_count', 0);
+    const entityValue = get(item, 'key', 0);
```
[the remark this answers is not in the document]
yizheliu-amazon: sure.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
yizheliu-amazon: If detector is HC detector, loading raw data is necessary for rendering anomaly result table like [anomalyResult](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/public/pages/DetectorResults/containers/AnomalyHistory.tsx#L408) which is from `atomicAnomalyResults`. Thus, in that case we need to have `showLoading` to be true so that UI can have loading status. 

In case of single entity detector, loading raw data is not necessary for rendering any UI result, because it is only used for feature missing checking. Thus, we don't need to set `showLoading` to be true in that case.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -99,58 +118,70 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
...
-              dateRange.endDate,
-              1,
-              props.detector.id
-            )
+      const result = await dispatch(
+        searchResults(
+          getBucketizedAnomalyResultsQuery(
+            dateRange.startDate,
+            dateRange.endDate,
+            1,
```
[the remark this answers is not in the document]
yizheliu-amazon: I think the `interval` parameter name in `getBucketizedAnomalyResultsQuery()` is kind of confusing. This is not related to detector interval, and more about bucket interval base, which is 1 minute. In [getBucketizedAnomalyResultsQuery()](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/pages/utils/anomalyResultUtils.ts#L362), we split the original time range(endTime - startTime) into MAX_DATA_POINTS(1000) buckets, and since time range is at millisecond level, then we divide by 1 minute * MIN_IN_MILLI_SECS, so that the result value is based on minute level, and then it can be used for `fixed_interval` sent to ES in [code](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/pages/utils/anomalyResultUtils.ts#L389).

After 2nd thought, I think the `interval` parameter can be removed.

public/pages/utils/anomalyResultUtils.ts
yizheliu-amazon: I think above [comment](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/350#discussion_r550337962) can explain it.

public/pages/AnomalyCharts/containers/AnomaliesChart.tsx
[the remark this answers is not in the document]
kaituo: got it.  Thanks for the explanation.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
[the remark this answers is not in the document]
kaituo: Does that mean if users select last 1 day's data and the detector interval is 1 hour, I still try to get 1000 buckets instead of 24 buckets?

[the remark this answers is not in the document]
yizheliu-amazon: no. This `interval` is not related to detector interval. For example: given a large date range selected, such as 2 weeks, which is 2 * 7 * 24 * 3600 * 1000 milliseconds(endTime - startTime), and we want to only show MAX_DATA_POINTS(1000) data points on chart, then we split above date range in milliseconds into 1000 buckets, then each bucket has `fixed_interval` as (2 * 7 * 24 * 3600 * 1000)/ 1000 = 1209600ms. Since we use [minute as unit](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/pages/utils/anomalyResultUtils.ts#L389) in query to ES, we divide it by 60(sec) * 1000(ms), aka 1209600/(60 * 1000) and then the result is the `fixed_interval` value. This is what `getBucketizedAnomalyResultsQuery()` does.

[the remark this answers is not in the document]
kaituo: I still don't get it.  Why doesn't getBucketizedAnomalyResultsQuery consider both data range and detector interval?  If there less than MAX_DATA_POINTS anomaly results to show, why do we still query for MAX_DATA_POINTS intervals?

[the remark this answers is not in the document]
kaituo: Discussed with Yizhe offline.  So we only call this method when there are more than MAX_DATA_POINTS anomaly results to show.

kaituo (approved):

PR #351, opened 2020-12-21
Add historical detector details page

ohltyler:
*Issue #, if available:*

*Description of changes:*

**This PR is 3 of 4 related to adding the historical detectors to the plugin. Once all 4 PRs are merged into the development branch, there will be a follow up PR to merge the development branch into master.**

This PR adds the historical detector details/results page (and related UT).

Screenshots:

Detector running with partial results:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-17-15_00_26 (1)

Detector finished:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-06-12_48_48

Edit detector modal:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-06-12_56_00

Delete detector modal:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-11-25-16_53_59

Delete detector modal (detector currently running):
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-06-12_56_18

### DIFF
--- a/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
+++ b/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
@@ -84,9 +84,13 @@ export const HistoricalDetectorConfig = (
         <EuiFlexItem>
           {renderCell(
             'Date range',
-            renderDate(get(props, 'detector.detectionDateRange.startTime', 0)) +
+            renderDate(
+              get(props, 'detector.detectionDateRange.startTime', undefined)
+            ) +
               '-' +
-              renderDate(get(props, 'detector.detectionDateRange.endTime', 0))
+              renderDate(
+                get(props, 'detector.detectionDateRange.endTime', undefined)
+              )
           )}
         </EuiFlexItem>
         <EuiFlexItem>
@@ -95,7 +99,7 @@ export const HistoricalDetectorConfig = (
         <EuiFlexItem>
           {renderCell(
             'Last updated',
-            renderDate(get(props, 'detector.lastUpdateTime', 0))
+            renderDate(get(props, 'detector.lastUpdateTime', undefined))
           )}
         </EuiFlexItem>
         <EuiFlexItem>
--- a/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
+++ b/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import ContentPanel from '../../../../components/ContentPanel/ContentPanel';
+import {
+  //@ts-ignore
+  EuiFlexGrid,
+  EuiFlexItem,
+  EuiText,
+  EuiFormRow,
+  EuiButton,
+  EuiFormRowProps,
+} from '@elastic/eui';
+import { Detector } from '../../../../models/interfaces';
+import React from 'react';
+import moment from 'moment';
+import { get } from 'lodash';
+
+interface HistoricalDetectorConfigProps {
+  detector: Detector;
+  isStoppingDetector: boolean;
+  onEditDetector(): void;
+}
+
+const FixedWidthRow = (props: EuiFormRowProps) => (
+  <EuiFormRow {...props} style={{ width: '250px' }} />
+);
+
+const renderCell = (title: string, description: string | number) => {
+  return (
+    <FixedWidthRow label={title}>
+      <EuiText>
+        <p className="enabled">{description}</p>
+      </EuiText>
+    </FixedWidthRow>
+  );
+};
+
+const renderDate = (obj: any) => {
+  if (typeof obj == 'number') {
+    return moment(obj).format('MM/DD/YYYY hh:mm A');
+  } else {
+    return '-';
+  }
+};
+
+export const HistoricalDetectorConfig = (
+  props: HistoricalDetectorConfigProps
+) => {
+  return (
+    <ContentPanel
+      title="Historical detector configuration"
+      titleSize="s"
+      panelStyles={{ margin: '0px' }}
+      actions={[
+        <EuiButton
+          data-test-subj="editDetectorButton"
+          onClick={props.onEditDetector}
+          disabled={props.isStoppingDetector}
+        >
+          Edit
+        </EuiButton>,
+      ]}
+    >
+      <EuiFlexGrid columns={3} gutterSize="l" style={{ border: 'none' }}>
+        <EuiFlexItem>
+          {renderCell('Name', get(props, 'detector.name', ''))}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell('ID', get(props, 'detector.id', ''))}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell(
+            'Date range',
+            renderDate(get(props, 'detector.detectionDateRange.startTime', 0)) +
+              '-' +
+              renderDate(get(props, 'detector.detectionDateRange.endTime', 0))
+          )}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell('Description', get(props, 'detector.description', ''))}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell(
+            'Last updated',
+            renderDate(get(props, 'detector.lastUpdateTime', 0))
+          )}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell(
+            'Data source index',
+            get(props, 'detector.indices.0', '')
+          )}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell(
+            'Detection interval',
+            get(props, 'detector.detectionInterval.period.interval', '') +
+              ' ' +
+              get(props, 'detector.detectionInterval.period.unit', '')
+          )}
+        </EuiFlexItem>
+      </EuiFlexGrid>
+    </ContentPanel>
+  );
+};
--- a/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorControls/HistoricalDetectorControls.tsx
+++ b/public/pages/HistoricalDetectorDetail/components/HistoricalDetectorControls/HistoricalDetectorControls.tsx
@@ -82,25 +82,18 @@ export const HistoricalDetectorControls = (
         </EuiPopover>
       </EuiFlexItem>
       <EuiFlexItem grow={false} style={{ marginLeft: '0px' }}>
-        {props.isStoppingDetector ? (
-          <EuiButton
-            data-test-subj="isStoppingDetectorButton"
-            onClick={() => {}}
-            isLoading={true}
-            iconType={'play'}
-          >
-            Stopping
-          </EuiButton>
-        ) : props.detector?.curState === DETECTOR_STATE.INIT ||
-          props.detector?.curState === DETECTOR_STATE.RUNNING ? (
+        {props.isStoppingDetector ||
+        props.detector?.curState === DETECTOR_STATE.INIT ||
+        props.detector?.curState === DETECTOR_STATE.RUNNING ? (
           <EuiButton
             data-test-subj="stopDetectorButton"
             onClick={() =>
               props.onStopDetector(HISTORICAL_DETECTOR_ACTION.STOP, undefined)
             }
+            isLoading={props.isStoppingDetector}
             iconType={'stop'}
           >
-            Stop historical detector
+            {props.isStoppingDetector ? 'Stopping' : 'Stop historical detector'}
           </EuiButton>
         ) : (
           <EuiButton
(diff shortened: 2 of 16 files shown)


public/pages/HistoricalDetectorDetail/utils/helpers.tsx
```diff
@@ -0,0 +1,136 @@
...
+          }
+          color="primary"
+        />
+      );
+    default:
+      return null;
+  }
+};
+
+export const getRunningHistoricalDetectorCount = (
```
ylwu-amzn, line 128: This function is not used ?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+        1
+      );
+      const historicalDetectorsResponse = await dispatch(
+        getHistoricalDetectorList(GET_ALL_DETECTORS_QUERY_PARAMS)
+      );
+      const runningCount = getRunningHistoricalDetectorCount(
+        get(historicalDetectorsResponse, 'response.detectorList')
+      );
+
+      if (nodeCount * 2 <= runningCount) {
```
ylwu-amzn, line 191: Max running historical detector per node is a dynamic setting. Default value is 2, but user can change it to other values like 10. The start detector API will check if task executing exceeds limit or not.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+  const isDark = darkModeEnabled();
+  useHideSideNavBar(true, false);
+  const dispatch = useDispatch();
+  const detectorId: string = get(props, 'match.params.detectorId', '');
+
+  const adState = useSelector((state: AppState) => state.ad);
+  const allDetectors = adState.detectors;
+  const errorGettingDetectors = adState.errorMessage;
+  const detector = allDetectors[detectorId];
+  const monitors = useSelector((state: AppState) => state.alerting.monitors);
```
ylwu-amzn, line 85: We can't create monitor for historical detector. Why need to get monitors here?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+  }, [detector]);
+
+  const fetchDetector = async () => {
+    try {
+      await dispatch(getDetector(detectorId));
+      setIsLoadingDetector(false);
+    } catch {
+      core.notifications.toasts.addDanger(
+        'Error retrieving the historical detector'
+      );
```
ylwu-amzn, line 128: If can't retrieve historical detector, should we call `setIsLoadingDetector(true)`  and redirect to historical detectors list?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+  // Try to get the detector initially
+  useEffect(() => {
+    if (detectorId) {
+      setIsLoadingDetector(true);
+      fetchDetector();
+    }
+  }, []);
+
+  // If detector is initialiazing or running: keep fetching every second to quickly update state/results/percentage bar, etc.
```
ylwu-amzn, line 140: Seems this description can't match the code, `HISTORICAL_DETECTOR_RESULT_REFRESH_RATE ` is 10 seconds

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+import { GET_ALL_DETECTORS_QUERY_PARAMS } from '../../../utils/constants';
+
+export interface HistoricalDetectorRouterProps {
+  detectorId?: string;
+}
+interface HistoricalDetectorDetailProps
+  extends RouteComponentProps<HistoricalDetectorRouterProps> {}
+
+interface HistoricalDetectorDetailModalState {
+  isOpen: boolean;
```
ylwu-amzn, line 68: What does `isOpen` mean?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+    } catch (err) {
+      core.notifications.toasts.addDanger(
+        `There was a problem starting the historical detector: ${err}`
+      );
+    }
+  };
+
+  // We query the task state 5s after making the stop detector call. If the task is still running,
+  // then it is assumed there was an error stopping this task / historical detector.
+  const onStopDetector = async (isDelete: boolean, listener?: Listener) => {
```
ylwu-amzn, line 209: `isDelete` seems not so readable, how about we use `HISTORICAL_DETECTOR_ACTION` here ? If need to stop detector for delete action, we can pass `HISTORICAL_DETECTOR_ACTION.DELETE`

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+        throw 'please try again.';
+      }
+      if (!isDelete) {
+        core.notifications.toasts.addSuccess(
+          'Successfully stopped the historical detector'
+        );
+      }
+      if (listener) listener.onSuccess();
+    } catch (err) {
+      core.notifications.toasts.addDanger(
```
ylwu-amzn, line 226: Should we set is stopping detector as false if any error happen?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: I see. So to clarify, I can remove all logic related to error handling on the frontend for this, and instead just propagate the backend error message?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: Good catch. Left over from copied existing code.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: After looking at this, I realized we don't need any toast here at all, since it will be caught in the custom hook on line 102 [here](https://github.com/ohltyler/anomaly-detection-kibana-plugin/blob/add-details-page/public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx#L102)., where a toast is displayed and the user is redirected to the historical detectors list. Will remove this catch block altogether.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: Thanks, fixed.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: `isOpen` just means if the modal is open or not. Further down in this file is the `getHistoricalDetectorDetailModal` function with uses the modal state to determine what, if any, modal should currently render.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: I think the idea of passing a `HISTORICAL_DETECTOR_ACTION` here makes much more sense, will change to that.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: And seems we would want to call `setIsLoadingDetector(false)` here instead?

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: Good catch, will call `setIsStoppingDetector(false)` in a `finally` block here.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ohltyler: After looking further, the entire `isLoadingDetector` field isn't used at all. Will remove this line entirely.

public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
```diff
@@ -0,0 +1,118 @@
...
+              renderDate(get(props, 'detector.detectionDateRange.endTime', 0))
+          )}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell('Description', get(props, 'detector.description', ''))}
+        </EuiFlexItem>
+        <EuiFlexItem>
+          {renderCell(
+            'Last updated',
+            renderDate(get(props, 'detector.lastUpdateTime', 0))
```
ylwu-amzn, line 98: In normal case `lastUpdateTime ` should exist. If `lastUpdateTime ` not exist, will use 0 as default value, will `renderDate` function throw exception for 0 ?

public/pages/HistoricalDetectorDetail/components/HistoricalDetectorControls/HistoricalDetectorControls.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </EuiContextMenuItem>
+          </EuiContextMenuPanel>
+        </EuiPopover>
+      </EuiFlexItem>
+      <EuiFlexItem grow={false} style={{ marginLeft: '0px' }}>
+        {props.isStoppingDetector ? (
+          <EuiButton
+            data-test-subj="isStoppingDetectorButton"
+            onClick={() => {}}
+            isLoading={true}
```
ylwu-amzn, line 89: Is it possible to just reuse the original <EuiButton> just set `isLoading={props.isStoppingDetector}` ? I think the button will not be clickable if `isLoading=true`, so no need to worry the `onClick` function.

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
ylwu-amzn: Got it

public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
ohltyler: It will still work as `0` is a valid timestamp, and `moment(0)` will not throw an exception, but it will cause an invalid date to display. In fact, `renderDate` already checks if the `typeof` date passed is a `number`.

It's probably better to just pass `undefined` here, which will render `-` instead of the current implementation which renders an invalid date, to help avoid possible confusion.

public/pages/HistoricalDetectorDetail/components/HistoricalDetectorControls/HistoricalDetectorControls.tsx
ohltyler: That's a good point. Will change to just have the `startDetectorButton` and `stopDetectorButton`.

public/pages/HistoricalDetectorDetail/components/HistoricalDetectorConfig/HistoricalDetectorConfig.tsx
```diff
@@ -84,9 +84,13 @@ export const HistoricalDetectorConfig = (
         <EuiFlexItem>
           {renderCell(
             'Date range',
-            renderDate(get(props, 'detector.detectionDateRange.startTime', 0)) +
+            renderDate(
+              get(props, 'detector.detectionDateRange.startTime', undefined)
```
ylwu-amzn, line 88: minor: I think `get` will return undefined by default if `detector.detectionDateRange.startTime` is missing.

ylwu-amzn (approved):

public/pages/HistoricalDetectorDetail/containers/HistoricalDetectorDetail/HistoricalDetectorDetail.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+  const onStartDetector = async () => {
+    try {
+      await dispatch(startDetector(detectorId));
+      core.notifications.toasts.addSuccess(
+        `Successfully started the historical detector`
+      );
+    } catch (err) {
+      core.notifications.toasts.addDanger(
+        `There was a problem starting the historical detector ${detector?.name}: ${err}`
```
yizheliu-amazon, line 171: may consider using `prettifyErrorMessage()` like [this](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/ec42b9fa6b62c7f7158b1178535bbcf5e519dbc6/public/pages/createDetector/containers/CreateDetector.tsx#L138-L140) to show prettier message in case of permission issue. Same comment for `err` case of StopDetector/DeleteDetector below

ohltyler: Thanks, agreed this is better. Will change for the existing historical detector actions.

ohltyler: Fixed in latest commit

yizheliu-amazon (approved):

PR #349, opened 2020-12-16
Add create / edit historical detector page

ohltyler:
*Issue #, if available:*

*Description of changes:*

**This PR is 2 of 4 related to adding the historical detectors to the plugin. Once all 4 PRs are merged into the development branch, there will be a follow up PR to merge the development branch into master.**

This PR adds the create/edit historical detector page (and related UT).

Screenshots:

Create page (using existing detector):
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-10-10_42_22

Edit page:
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-10-10_43_34

Callout (no existing detector with selected index):
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-10-10_44_27

Callout (no index selected):
screencapture-localhost-5601-app-opendistro-anomaly-detection-kibana-2020-12-10-10_44_41

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #352, opened 2020-12-22
Add other changes related to historical workbench

ohltyler:
*Issue #, if available:*

*Description of changes:*

**This PR is 4 of 4 related to adding the historical detectors to the plugin. Once all 4 PRs are merged into the development branch, there will be a follow up PR to merge the development branch into master.**

This PR adds several changes to existing pages, server-side code, redux store, and data models to accommodate the addition of the historical detector workbench.

Changes include:
- Filtering out historical detectors on existing real-time pages (dashboard and detector list)
- Adding side nav hierarchy and the additional historical detector list page
- Wording changes proposed by the tech writer that affected both real-time and historical portions of the plugin
- Minor formatting changes proposed by UX reviewers that affected both real-time and historical portions of the plugin
- Updating and adding some data models
- Adding an `isNotSample` prop and historical-detector-related props to various chart components to differentiate between real-time results, sample results, and historical results
- Adds historical detector list in redux store and various functions to retrieve them
- Adds some checks in `getDetector()` server function to retrieve task-related info if the detector is historical
- Adds a `Finished` detector state that is only applicable to historical detectors
- Adds and removes some stats in the anomaly history chart, also removes the confidence line chart entirely for the historical detector case

Screenshot of updated anomaly history chart for the historical case.
1. Removes confidence and last anomaly occurrence fields,
2. Removes confidence in line chart,
3. Adds avg anomaly grade stat

Screen Shot 2020-12-28 at 12 36 19 PM


public/pages/DetectorResults/containers/AnomalyHistory.tsx
```diff
@@ -343,7 +370,7 @@ export const AnomalyHistory = (props: AnomalyHistoryProps) => {
         {isLoading || isLoadingAnomalyResults ? (
           <EuiFlexGroup
             justifyContent="spaceAround"
-            style={{ height: '200px', paddingTop: '100px' }}
+            style={{ height: '500px', paddingTop: '100px' }}
```
ylwu-amzn, line 373: Why change to 500px here ? Will it impact realtime detector?

public/utils/utils.tsx
```diff
@@ -163,3 +165,17 @@ export const getTitleWithCount = (title: string, count: number | string) => {
     </EuiTitle>
   );
 };
+
+export function convertTimestampToString(timestamp: number | string) {
+  if (typeof timestamp === 'string') {
+    return timestamp;
+  }
+  return moment(timestamp).format();
```
ylwu-amzn, line 173: Should we add date format here ? Like "MM/DD/YYYY hh:mm A"

public/utils/utils.tsx
```diff
@@ -163,3 +165,17 @@ export const getTitleWithCount = (title: string, count: number | string) => {
...
+}
+
+export function convertTimestampToNumber(timestamp: number | string) {
```
ylwu-amzn, line 176: This method is not being used currently ?

server/routes/ad.ts
```diff
@@ -220,38 +233,88 @@ export default class AdService {
         .callAsCurrentUser('ad.getDetector', {
           detectorId,
         });
+
+      // Considering any detector with no detection date range to be a realtime detector
```
ylwu-amzn, line 237: Get detector API will also return detector type https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorType.java#L21

I'm ok to use either detection date range or this type field.

server/routes/ad.ts
```diff
@@ -460,8 +523,8 @@ export default class AdService {
         mustQueries.push({
           query_string: {
             fields: ['indices'],
-            default_operator: 'OR',
-            query: `*${indices.trim().split(' ').join('* *')}*`,
+            default_operator: 'AND',
```
ylwu-amzn, line 526: Why need to change from OR to AND here ?

server/routes/utils/adHelpers.ts
```diff
@@ -253,3 +253,93 @@ export const getErrorMessage = (err: any) => {
...
+    }
+  });
+  return detectorMapWithTaskInfo;
+};
+
+// Four checks/transformations need to be made here:
+// (1) set to DISABLED if there is no existing task for this detector
+// (2) set to UNEXPECTED_FAILURE if the task is in a FAILED state to stay consistent
+// (3) set to INIT if the task is in a CREATED state
+// (4) set to DISABLED if the task is in a STOPPED state
```
ylwu-amzn, line 332: Does not cover all stats , check all task states https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java#L19.

INIT, RUNNING and FINISHED can be used directly like realtime detector?

public/models/interfaces.ts
```diff
@@ -184,7 +197,20 @@ export type AnomalySummary = {
   lastAnomalyOccurrence: string;
 };
 
+export type HistoricalDetectorAnomalySummary = {
```
yizheliu-amazon, line 200: Is this being used? Looks like [AnomalySummary](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/14168a737f361203d0b2434006b04d55c65c0dfd/public/models/interfaces.ts#L178) can be reused.

public/models/interfaces.ts
```diff
@@ -184,7 +197,20 @@ export type AnomalySummary = {
...
+  minConfidence: number;
+  maxConfidence: number;
+};
+
 export type DateRange = {
   startDate: number;
   endDate: number;
 };
+
+export type DetectionDateRange = {
```
yizheliu-amazon, line 213: I think re-using `DateRange` should be sufficient.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
ohltyler: This change was to prevent the page to look as jerky when in the loading state, since it would often cause the screen to add & remove the scrollbar. It does affect the realtime detector, but improves that as well since it is just an aesthetic improvement.

public/utils/utils.tsx
ohltyler: Good catch, will add the enforced output format.

public/utils/utils.tsx
ohltyler: It is actually being used in the create historical detector page when converting Formik values to the numbers when making a create/update detector request. It's not included in this PR, but you can see it used [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/349/files)

server/routes/ad.ts
ohltyler: I see, thanks. I think I will leave as-is for now, to prevent adding more frontend data types and to keep consistent with the alerting changes.

server/routes/ad.ts
ohltyler: This is related to returning correct results when an `indices` search field is passed to the `getDetectors()` function. Before, there was no calls that were actually utilizing this field so the bug was ignored, but it is now used in the create detector page when finding existing detectors based on the selected index.. Note that the above `search` field uses the correct `AND` operator as well.

server/routes/utils/adHelpers.ts
ohltyler: So the 4 checks are just to transform any of the returned historical detector states to match up with the states recognized by the frontend (the `DETECTOR_STATE` enum [here](server/utils/constants.ts)). And yes, if the state is `INIT`/`RUNNING`/`FINISHED`, since those are all recognized frontend states, no transformation is needed.

I decided to combine the real-time and historical states into one enum because of the end goal of having both detector types be a more singular data model.

Please let me know if you have any other suggestions here regarding the changes here. My reasoning for not exposing the `CREATED` state, and instead encapsulating it as part of the `INIT` state, was just to simplify the frontend changes and not add too many new terms and states for the user to have to understand.

public/models/interfaces.ts
ohltyler: This is a good catch. And actually, the reason for having a separate data model here is because the `lastAnomalyOccurrence` field is not applicable for the historical detector case, and is not part of the mockups.

Looks like currently, the `HistoricalDetectorAnomalySummary` type is not being used properly in the `AnomalyHistory` component. I will need to add a check to show the appropriate summary based on the detector type.

ohltyler: After looking more, seems it will be easier to not add a new `HistoricalDetectorAnomalySummary` here at all like you suggested, and instead just remove the occurrence item if render historical detector results. Will remove  `HistoricalDetectorAnomalySummary`.

server/routes/ad.ts
ylwu-amzn: Sure, sounds good

server/routes/ad.ts
```diff
@@ -460,8 +523,8 @@ export default class AdService {
...
+            query: `*${indices.trim().split('-').join('* *')}*`,
```
ylwu-amzn, line 527: Will it work just splitting by "-" ? Have you tested with " "

server/routes/ad.ts
ylwu-amzn: Thanks for fixing this old bug

server/routes/utils/adHelpers.ts
ylwu-amzn: Sounds good. `CREATED` state will be very short-lived for normal case, it will change to `INIT` very quickly(<1s), so no need to show this state on frontend.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
ylwu-amzn: Got it, will it impact the realtime detector's AD result chart?

public/models/interfaces.ts
ohltyler: Per the mockup, have removed the confidence showing on the chart entirely for the historical detector case.

ohltyler:
Regarding the changes for the anomaly history chart, please see the added screenshot in the description and the latest commit.

public/pages/DetectorResults/containers/AnomalyHistory.tsx
ohltyler: Yes, please see above comment. It's an improvement for both real-time and historical.

public/utils/utils.tsx
ohltyler: Fixed in latest commit.

public/models/interfaces.ts
ohltyler: The reason for adding this `DetectionDateRange` is because the fields understood by the backend take `startTime` and `endTime` fields instead of `startDate` and `endDate` fields used by the `DateRange`. I didn't want to add more converting when parsing existing detectors or creating/updating new detectors. Let me know what you think.

server/routes/ad.ts
ohltyler: Yes, this is another part of the bug that needed to be fixed. Note that the `search` field also splits by `-` in order to properly search for detectors with a `-` in them. A similar fix needs to be done for detectors that are using indices with `-` in them. See [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/pull/272#discussion_r472553954) for more details.

yizheliu-amazon (approved):

ylwu-amzn (approved):

Issue #342, opened 2020-12-08
Can't add feature for the detector when users use AD first time


weicongs-amazon:
### ODFE Version ###
1.12.0

### Distributions ###
All: Docker, RPM, Tar, Deb

### Issue Description ###
After one new ODFE cluster is set up, customer uses the kibana AD UI to create a new detector. The detector can be created successfully, but when customer tries to add new feature, and it will always fail with error message
```
Cannot create anomaly detector with name [test-1] as it's already used by detector [u4Z5RHYBNpJRqK5Wectv]
```

After this detector is deleted, everything works well with new detectors.
### How to reproduce it ###
* Create a new ODFE cluster
* Add kibana sample log data
* create a new detector
* add a new feature

### Isse Severity ###
Not urgent since customer can delete the first detector and then everything works well

ohltyler:
Fixed by #353, closing.

PR #353, opened 2020-12-24
Fix failure of adding feature to 1st detector in cluster

yizheliu-amazon:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/342
*Description of changes:*
When creating 1st detector, adding feature fails because `seqNo` is 0 for the 1st created detector, and in our code, we use `if (ifSeqNo && ifPrimaryTerm)` to determine if we should update detector or create detector. In Javascript, 0 is considered as `false`, that's why there is such issue. Changing to use `isNumber()` fixs the issue. 

### DIFF
--- a/server/routes/ad.ts
+++ b/server/routes/ad.ts
@@ -46,7 +46,7 @@ import {
   isIndexNotFoundError,
   getErrorMessage,
 } from './utils/adHelpers';
-import { set } from 'lodash';
+import { isNumber, set } from 'lodash';
 import {
   RequestHandlerContext,
   KibanaRequest,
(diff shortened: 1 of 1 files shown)


ylwu-amzn (approved):

ylwu-amzn:
Please test by starting local cluster and test both first realtime and HC detectors.

yizheliu-amazon:
> Please test by starting local cluster and test both first realtime and HC detectors before merging.

Yeah. Already tested both cases.

ohltyler (approved):

PR #357, opened 2021-01-15
Fix zoom bug and remove confidence from historical results table

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR adds some bug fixes & improvements to the historical detectors pages. Specifically:

1. Fixes the bug of showing empty results if zooming in from bucketized anomaly results -> raw anomaly results (unrelated to historical detectors specifically)
2. Removes the confidence from anomaly results table in the historical detector detail page
3. Moves the sample detectors link in the side nav under the Real-time detectors hierarchical component.

Confirmed all UT pass.

Screenshots of changes:

Screen Shot 2021-01-15 at 10 27 01 AM

Screen Shot 2021-01-15 at 10 27 18 AM


ylwu-amzn (approved):
LGTM. Thanks for the change!

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -104,11 +104,9 @@ export const buildParamsForGetAnomalyResultsWithDateRange = (
...
     sortDirection: SORT_DIRECTION.DESC,
     sortField: AD_DOC_FIELDS.DATA_START_TIME,
-    dateRangeFilter: JSON.stringify({
-      startTime: startTime,
-      endTime: endTime,
-      fieldName: AD_DOC_FIELDS.DATA_START_TIME,
-    }),
+    startTime: startTime,
+    endTime: endTime,
+    fieldName: AD_DOC_FIELDS.DATA_START_TIME,
```
yizheliu-amazon, line 109: Thanks for the fix. It looks like [getDetectorLiveResults()](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/da5ead30a8266f43ff5212d59f18ec75f4559a8b/public/redux/reducers/liveAnomalyResults.ts#L62-L71) is calling same function [getAnomalyResults](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/historical-workbench-dev/server/routes/ad.ts#L732). I guess the current live result on Detector Result page is having issue. 

If yes, can you also make the change to [getQueryParamsForLiveAnomalyResults()](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/historical-workbench-dev/public/pages/utils/anomalyResultUtils.ts#L63-L81)?

ohltyler: I see. Yes, looks like `getQueryParamsForLiveAnomalyResults()` would have a similar issue as it is passing a `dateRangeFilter` prop and then calling `getAnomalyResults()` like you pointed out, good catch. I will test out a similar fix for this function.

What is the current issue that you are seeing on the live results component? Is it also related to zooming?

yizheliu-amazon: There is no issue on master branch because [dateRangeFilter](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/master/server/routes/ad.ts#L685) is not removed yet on master branch. 

I haven't played with [historical-workbench-dev](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/tree/historical-workbench-dev) in my local workspace, not sure what exact issue this can cause. In live detector result chart, we want to show anomalies in last 60 intervals, where 60 is the `size` specified in [query](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/blob/historical-workbench-dev/public/pages/utils/anomalyResultUtils.ts#L72). The only issue with dateRangeFilter or startTime missing I can think of is: we might return result older than 60 intervals if some of the latest 60 intervals have anomaly result missed

ohltyler: Got it. I've confirmed on my local that currently it will pass a `startTime` of 0 which is the default set in `getAnomalyResults()`. Since the query passed to ES is a `gte` based on the `startTime`, this will cause issues once there are a lot of results for the detector, where newer results will be ignored.

I've now fixed this, and confirmed that it will now pass the desired `startTime` when retrieving live results. Will push a new commit.

public/pages/utils/anomalyResultUtils.ts
```diff
@@ -99,6 +97,7 @@ export const buildParamsForGetAnomalyResultsWithDateRange = (
   anomalyOnly: boolean = false,
   entity: Entity | undefined = undefined
 ) => {
+  console.log('getting regular anomaly results');
```
yizheliu-amazon, line 100: unnecessary log.

yizheliu-amazon (approved):
Thanks for the fix.

public/pages/utils/anomalyResultUtils.ts
ohltyler: thanks, removed.

Issue #348, opened 2020-12-15
Overlap between anomaly bars in live anomaly chart


ylwu-amzn:
We will show live anomaly on AD detail page. If there are consecutive anomalies, the bars will have overlap. 

![Screen Shot 2020-12-15 at 9 58 11 AM](https://user-images.githubusercontent.com/49084640/102254222-38f12180-3ebd-11eb-84a5-ff02940ac61b.png)

ohltyler:
Looking into this.

PR #362, opened 2021-01-25
Fix live chart bar width

ohltyler:
*Issue #, if available:* #348

*Description of changes:*

This PR fixes the issue of overlapping annotations on the live anomalies chart in the detector results page, if there are anomalies found in consecutive intervals. The `RectAnnotation` component is limited in that a specific width cannot be set on it - to make up for this, we can use the `strokeWidth` to set the border size of the annotation, then set the stroke color to be consistent with the chart background to make the annotations appear smaller and not overlapping.

Confirmed all UT & IT pass. Confirmed it works consistent in dark mode.

Before:
Screen Shot 2021-01-25 at 9 13 54 AM

After:
Screen Shot 2021-01-25 at 9 12 04 AM


yizheliu-amazon (approved):
Thanks for the fix.

ylwu-amzn (approved):
LGTM. Thanks for the fix!

PR #363, opened 2021-01-25
Add sleep time before running Cypress tests

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR adds time for the plugin to come up and running before running the Cypress tests. The shorter time was causing the first tests to be flaky and occasionally fail in the GitHub runners due to the plugin not being ready.

Example of a passing run: https://github.com/ohltyler/anomaly-detection-kibana-plugin/actions/runs/510110841


yizheliu-amazon (approved):
thanks for the fix.

ylwu-amzn (approved):

Issue #361, opened 2021-01-23
historical detector: show readable error message when there is no historical data


ylwu-amzn:
When there is no data in historical detection date range, historical detector task will throw ResourceNotFoundException and write such error message in AD task: `There is no data in the detection date range` . But we will show `Unexpected failure ` on historical detector detail page. Suggest to show a more readable error message

ohltyler:
Taking a look. Having discussions on the idea of adding an additional 'No data' state (or something similar) for historical detectors.

PR #364, opened 2021-01-25
Add 'No data' state for historical detectors

ohltyler:
*Issue #, if available:* #361

*Description of changes:*

This PR adds a 'No data' state to the historical detectors, to make it more clear to the user when a historical detector's detection date range doesn't return any results / no data to aggregate over. Previously, this error was captured under the vague 'Unexpected failure' detector state, and provided no actionable steps for the user to try to fix it.

Specific changes:
1. Adds the 'No data' state to the list of detector states, only displays specifically for historical detectors
2. Adds a callout with a button to edit the historical detector when in the 'No data' state
3. Adds corresponding UT

Screenshots:
image

image


ohltyler:
Will confirm with UX on the changes before merging.

server/routes/utils/adHelpers.ts
```diff
@@ -325,15 +325,18 @@ export const appendTaskInfo = (
   return detectorMapWithTaskInfo;
 };
 
-// Four checks/transformations need to be made here:
+// Five checks/transformations need to be made here:
```
yizheliu-amazon, line 328: minor: change `Five` to `Following`, so that we don't need to change it every time we have new checks.

yizheliu-amazon (approved):

server/routes/utils/adHelpers.ts
ohltyler: Agreed, changed.

server/utils/constants.ts
```diff
@@ -69,6 +69,7 @@ export enum DETECTOR_STATE {
   FEATURE_REQUIRED = 'Feature required',
   INIT_FAILURE = 'Initialization failure',
   UNEXPECTED_FAILURE = 'Unexpected failure',
+  NO_DATA = 'No data',
```
ylwu-amzn, line 72: Not related with this PR. But we may need to test more edge cases like not enough data, exceed our max running AD task limit etc.

ylwu-amzn (approved):

server/utils/constants.ts
ohltyler: Yeah, agreed. I think this "no data" case can be pretty common, so definitely good to have it separated here, but definitely doesn't cover all cases that would currently fall under the 'Unexpected failure' state. I'm assuming those other cases would fall under the `FAILED` state from the backend, and it's probably a matter of parsing out the error message to categorize into other states similar to what's done here.

ylwu-amzn: Yes, we can also parse the error message for "FAILED" state. But not so easy for exceptions which has no readable error message.

Issue #372, opened 2021-02-02
Dashboard not showing empty state when no real-time detectors


ohltyler:
Currently, if there are historical detectors created but no real-time detectors, the dashboard does not show the empty state like it should. Change so the dashboard only looks at the count or real-time detectors instead of real-time + historical.

Issue #304, opened 2020-09-03
Improve loading state of dashboard page


ohltyler:
Currently, when refreshing the page on the dashboard, it will quickly render the empty dashboard page before loading the full dashboard page (if detectors exist). This is because the page will render before retrieving all detectors (if any) from the backend, so it is assuming there are no existing detectors. 

This can cause some user confusion to see this flicker of no detectors even if they have detectors created. And, if it takes a long time to retrieve the detectors (cluster under heavy load, low system resources, etc.), this empty dashboard page may show for an extended amount of time, which would be very confusing.

Change to add an intermediate loading state for the dashboard, if detectors are still being retrieved from the backend.

PR #373, opened 2021-02-02
Fix dashboard loading state and empty state logic

ohltyler:
*Issue #, if available:*

*Description of changes:*

This PR fixes 2 issues:
1. The loading state, which would occasionally flash an incorrect empty state while the detectors are being retrieved from ES (issue #304).
2. The empty state, which was not showing up when no real-time detectors existed, but historical detectors did (issue #372).

Confirmed all UT & IT pass.


yizheliu-amazon (commented):
thanks for the fix

ylwu-amzn (approved):

yizheliu-amazon (approved):

